PR #67, opened 2023-02-07
PAR halo region filling to fix #66

### LINKED ISSUE #66: PAR interpolation incorrect above top grid point
@syou83syou83 has noticed that interpolating the PAR field at a point above the top center point gives a value lower than that at the top point. This is probably because we've not been filling the halo region so its interpolating between [top value] and 0, the default for the halo region. 

This should be easily fixed by filling the top halo points with the surface value, or surface ^ 2 / top point so that interpoalting to the surface will give the correct value. I will do this tomorrow.

jagoosw:
Fixes #66


jagoosw:
Not sure why the test workflow isn't running, I've run most locally and this shouldn't effect them anyway so going to merge so Si can use it

PR #68, opened 2023-02-27
Better sediment integration

jagoosw:
Complete refactor of how sediment is coupled and integrated. Here I have focused on minimusing the amount of Oceananigans functions I'm having to replace, as it is I think the minimum possible is to only overload the `ab2_step!` and equivilant functions.

Not yet working but thought I should post progress anyway.


jagoosw:
Timestepping now occures, even if it does go NaN after about 2 steps

PR #69, opened 2023-02-27
Better sediment integration

jagoosw:
Complete refactor of how sediment is coupled and integrated. Here I have focused on minimusing the amount of Oceananigans functions I'm having to replace, as it is I think the minimum possible is to only overload the ab2_step! and equivilant functions. Everything else is done through the `update_tendencies!` in the new bgc interface.

Not yet working but thought I should post progress anyway.

Timestepping now "works", even if it does go NaN after a few steps.


jagoosw:
test
test

Looks to be getting steady state after 2 years, will run for longer to see if it equilibriates.

I have no idea if these results are physical but I think the model is working as specified in Soetaert et al. 2000. I will try to find some model results to compare to but I'm fairly happy with this as a proof of concept for anyone else who might want to use a sediment model.

jagoosw:
To do:

- [ ] Add RK3 timestepper
- [ ] (Find some results to compare to)
- [ ] Write docstrings
- [ ] Update docs pages

jagoosw:
Completed GPU fixes here so superceeds #71

jagoosw:
Also carried on running this for much longer and it looks like it converges nicely just very very slowly
test

PR #92, opened 2023-05-17
Docs were using a particular commit of OceanBioME.jl

navidcy:
Since the docs built gh action uses `julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'` there is no need to include a docs/Manifest file. If we want to ensure particular dependency versions for building the docs then we must not use `Pkg.develop()`.

But even if we choose to use a Manifest file, at the moment the docs/Manifests points to a local repo clone,

https://github.com/OceanBioME/OceanBioME.jl/blob/3378d6df3b2d463a4b4a5d8a3ec0c073525fbde5/docs/Manifest.toml#L1116

which will definitely won't work.


jagoosw (approved):
Looks good, thanks!

PR #85, opened 2023-04-12
Better latex rendering in docs

navidcy:

PR #93, opened 2023-05-17
Minor docs enhancements

navidcy:
This PR supercedes #85.


jagoosw (approved):
Thank you!

jagoosw:
Thank you for all this! Are you happy for me to merge this and #92? I think you would normally be able to but because it's not run the tests you can't right?

navidcy:
> Thank you for all this! Are you happy for me to merge this and #92? I think you would normally be able to but because it's not run the tests you can't right?

Sure. Merge!
Why didn’t the tests run though?

jagoosw:
> > Thank you for all this! Are you happy for me to merge this and #92? I think you would normally be able to but because it's not run the tests you can't right?
> 
> Sure. Merge! Why didn’t the tests run though?

They're set to only run when `src` or `test` change. Its from when the repo was private so we had limited action minutes but I guess I may as well change it to run for all changes now

PR #99, opened 2023-05-25
Simplify plotting in `README` example

navidcy:
I get the same movie.


navidcy:
I think that something was messed up with the new `gh-pages` branch while this PR was open. Closing in favor of #100.

PR #101, opened 2023-05-26
Add default parameters table to docs

### LINKED ISSUE #98: `SLatissima` parameters do not show
This snipped

https://github.com/OceanBioME/OceanBioME.jl/blob/1c5598ec482c97f92008250d6071c3d491a3e5c1/docs/src/appendix/params/SLatissima.md#L4-L5

doesn't seem to be doing anything.

jagoosw:
This adds the automatic generation of pages showing the default parameter values for models. I think this is quite important because I've often struggled trying to find parameter values for BGC models because often a couple are missed in papers, but with this we know the defaults in some version and can just detail variations from default.

Fixes #98


docs/src/appendix/params/index.md
```diff
@@ -1,5 +1,4 @@
 # Parameters 
-
 These pages contain most of the parameters used in the models included in OceanBioME.jl for ease of reference.
 
-At some point I will write a script that presents them in a better way.
+Units are generally `mmol X / m³` for concentration, and SI otherwise.gi
```
navidcy, line 4: .gi typo?

navidcy (commented):
Let’s see the preview before approving

navidcy (commented):
Let’s see the preview before approving

docs/src/appendix/params/index.md
jagoosw: Yeah, will fix

jagoosw:
> Let’s see the preview before approving

Yeah, seems to work locally but lets see

jagoosw:
Not sure whats caused this docs error, other people seem to have been getting the same recently: https://github.com/trixi-framework/Trixi.jl/issues/1475

navidcy:
I can dig in. I think I was getting this yesterday snd fixed it. It because eg there are two doc sections with name LOBSTER so `[LOBSTER](@ref) ` becomes ambiguous

jagoosw:
Ah I see, I though they had to be names like `(@id LOBSTER)` to be referenced. I imagine this is coming from the page titles and the new parameter pages so I will try changing the automatic title to `X Parameters` rather than just `X`.

navidcy:
This is one way to reference them. But also module or methods names. And eg because the LOBSTER constructor is in the docs library then if you make another page with @id LOBSTER you run into trouble. 🦞

jagoosw:
Ah I see, thank you

navidcy:
> This is one way to reference them. But also module or methods names. And eg because the LOBSTER constructor is in the docs library then if you make another page with @id LOBSTER you run into trouble. 🦞

(i think that’s the culprit)

navidcy (approved):
lgtm

it'd be nice if units appeared... but it's definitely good for now!

jagoosw:
> lgtm
> 
> 
> 
> it'd be nice if units appeared... but it's definitely good for now!

Yeah that would be good, I'm not sure how, maybe with something like Unitful.

PR #102, opened 2023-05-26
Fix Eady example + few more tweaks in other examples

navidcy:
I was used from having imposed U velocity and didn't notice that the example was imposing a V so I had flipped the plot so that  propagation is in x --- fixed it here! :)

(Plus some more plotting tweaks)


jagoosw (approved):
Thank you!

PR #106, opened 2023-05-28
Add `Docs/Visualization` page + convert figs to vector form (`svg`)

navidcy:
+ adjusts some figures that were plotting transposed arrays by accident
+ removes `interpolate = true` which allowed for mistakenly plotting rotated arrays without erroring.
+ adds some remarks about unit conversions in examples


navidcy:
Was there any particular reason `interpolate = true` kwarg was provided to `heatmap`?

jagoosw:
> Was there any particular reason `interpolate = true` kwarg was provided to `heatmap`?

Not really, I think it was just because it looks nicer sometimes

jagoosw:
Its not from this PR but I've just noticed that the air/sea CO2 flux and sinking flux are the same line in the column (and probably other) examples, I think because this line: https://github.com/OceanBioME/OceanBioME.jl/blob/e7bd802d853c8e5118bb4510506f1b8f81b0a1c2/examples/column.jl#L99

Doesn't create two separate arrays but references one with both variables. I think it has to be something like:
```julia
air_sea_CO₂_flux, carbon_export = zeros(length(times)),  zeros(length(times)) 
```

(they shouldn't be the exact same but average to the same over a long time.

navidcy:
> Its not from this PR but I've just noticed that the air/sea CO2 flux and sinking flux are the same line in the column (and probably other) examples, I think because this line:
> 
> https://github.com/OceanBioME/OceanBioME.jl/blob/e7bd802d853c8e5118bb4510506f1b8f81b0a1c2/examples/column.jl#L99
> 
> Doesn't create two separate arrays but references one with both variables. I think it has to be something like:
> 
> ```julia
> air_sea_CO₂_flux, carbon_export = zeros(length(times)),  zeros(length(times)) 
> ```
> 
> (they shouldn't be the exact same but average to the same over a long time because the air/sea exchange is slow)

You are correct!

```Julia
julia> a = b = zeros(3)
3-element Vector{Float64}:
 0.0
 0.0
 0.0

julia> for j in 1:3; a[j]=j; end

julia> a
3-element Vector{Float64}:
 1.0
 2.0
 3.0

julia> b
3-element Vector{Float64}:
 1.0
 2.0
 3.0
```

navidcy:
One this that is not explained is why each of these are computed at different depths? Also what depth this computation is made is never mentioned...

E.g. below `air_sea_CO₂_flux` at `k = 50` while `carbon_export` at `k = end-20`.

```Julia
    air_sea_CO₂_flux[i] = CO₂_flux.condition.parameters(0.0, 0.0, t, DIC[1, 1, 50, i], Alk[1, 1, 50, i], temp(1, 1, 0, t), 35)
    carbon_export[i] = (sPOM[1, 1, end-20, i] * model.biogeochemistry.sinking_velocities.sPOM.w[1, 1, end-20] +
                        bPOM[1, 1, end-20, i] * model.biogeochemistry.sinking_velocities.bPOM.w[1, 1, end-20]) * model.biogeochemistry.organic_redfield
```

navidcy:
Also this `* (12 + 16 * 2) * year / 1e6`: I get that is some sort of unit conversion but 12+16*2 does not ring any bells.

jagoosw:
I should probably add some description to this. Basically the air-sea flux is at the surface (maybe should change 50 to grid.Nz), and the sinking flux is assuming that anything that sinks below a certain depth is exported (but in this is only 100m I think, which is quite shallow).

The unit conversion is mmol N to kg CO2 so the 12 + 2x16 is for the mass of carbon dioxide

navidcy:
I added something. Feel free to modify/edit/push on the branch!

navidcy:
@jagoosw any idea why this happens?

https://github.com/OceanBioME/OceanBioME.jl/actions/runs/5115423233/jobs/9196670493?pr=106#step:5:725

jagoosw:
I'm not sure why this happens sometimes with the Eady example. That error is from the phytoplankton going negative but I'm not sure how that happens with:
```julia
simulation.callbacks[:neg] = Callback(scale_negative_tracers; callsite = UpdateStateCallsite())
simulation.callbacks[:nan_tendencies] = Callback(remove_NaN_tendencies!; callsite = TendencyCallsite())
simulation.callbacks[:abort_zeros] = Callback(zero_negative_tracers!; callsite = UpdateStateCallsite())
```

I'll re-run it and if that doesn't work try and work out what causes it locally.

johnryantaylor:
Lowering the CFL number is a good idea.  Another approach would be to update the CFL criterion every timestep instead of every 10 iterations as it is currently done.  The Eady problem is highly frontogenetic and the flow can change very rapidly - hence a timestep that was stable a few iterations ago might become too large.

glwagner:
> I can't get it to reliably run now even with a lower cfl, could it be the change to the diffusion?

Hmm not sure but we find that WENO typically provides sufficient horizontal dissipation, so you may want to eliminate explicit horizontal diffusion altogether. Most users shouldn't use Laplacian horizontal diffusion anyways for this kind of problem...

jagoosw:
Removing the horizontal diffusion makes it run again so I think it is that

jagoosw (approved):
All lgtm, thanks!

PR #109, opened 2023-06-20
small GPU bugfix

### LINKED ISSUE #75: GPU memory leak (?) when running with particles
As described in #73, I have a problem when I try to run a model with particles that the RAM usage increases until slurm kills the process (requesting more memory just makes it take longer to be killed) and never manages to complelte a timestep. I have also experianced this with normal Oceananigans particles so will investigate.

simone-silvestri:
I removed `model` from kernel calls as we do not have an `adapt_structure` method for models in Oceananigans (we do not want to pass the whole model to kernels)

Also, I switched the `adapt_structure` inside the overloading of `adapt_structure` to `adapt` calls as 

https://github.com/JuliaGPU/Adapt.jl/blob/01052e63decd251cdd4ba86d73327ad0de043a9f/src/Adapt.jl#L40
(I am not sure if it is better but I think this is the intended syntax)

Also I removed the 
```
NamedTuple{keys(lobster.sinking_velocities)}(ntuple(n -> adapt_structure(to, lobster.sinking_velocities[n]), length(lobster.sinking_velocities))), # not sure this is the most efficient way todo this
```
as 
https://github.com/JuliaGPU/Adapt.jl/blob/01052e63decd251cdd4ba86d73327ad0de043a9f/src/base.jl#L3
(also here, maybe not necessary but probably it has some benefits)

I tried `examples/kelp.jl` on our local V100s and it seems to work, @jagoosw let me know if this fixes #75 on your end!


examples/kelp.jl
```diff
@@ -37,10 +38,12 @@ nothing #hide
 @inline temp(x, y, z, t) = 2.4 * cos(t * 2π / year + 50day) + 10
 nothing #hide
 
+architecture = GPU()
+
```
navidcy, line 42: let's test it with GPU() but change to CPU() before merge?

src/Boundaries/Sediments/simple_multi_G.jl
```diff
@@ -185,9 +184,9 @@ sediment_fields(model::SimpleMultiG) = (C_slow = model.fields.C_slow, C_fast = m
...
-        model.timestepper.Gⁿ.NH₄[i, j, 1] += (Nᵐⁱⁿ * (1 - pₙᵢₜ)) / Δz
-        model.timestepper.Gⁿ.NO₃[i, j, 1] += (Nᵐⁱⁿ * pₙᵢₜ - Cᵐⁱⁿ * pᵈᵉⁿⁱᵗ * 4//5) / Δz
-        model.timestepper.Gⁿ.DIC[i, j, 1] += Cᵐⁱⁿ / Δz
-        model.timestepper.Gⁿ.O₂[i, j, 1]  -= max(0, (1 - pᵈᵉⁿⁱᵗ - pₐₙₒₓ * pₛₒₗᵢ) * Cᵐⁱⁿ / Δz) # this seems dodge but this model doesn't cope with anoxia properly
+        timestepper.Gⁿ.NH₄[i, j, 1] += (Nᵐⁱⁿ * (1 - pₙᵢₜ)) / Δz
+        timestepper.Gⁿ.NO₃[i, j, 1] += (Nᵐⁱⁿ * pₙᵢₜ - Cᵐⁱⁿ * pᵈᵉⁿⁱᵗ * 4//5) / Δz
+        timestepper.Gⁿ.DIC[i, j, 1] += Cᵐⁱⁿ / Δz
+        timestepper.Gⁿ.O₂[i, j, 1]  -= max(0, (1 - pᵈᵉⁿⁱᵗ - pₐₙₒₓ * pₛₒₗᵢ) * Cᵐⁱⁿ / Δz) # this seems dodge but this model doesn't cope with anoxia properly
     end
 end
```
navidcy, line 192:
```suggestion
end

```

jagoosw:
This example works for me on a P100 too. Thank you!

examples/kelp.jl
```diff
@@ -37,10 +38,12 @@ nothing #hide
 @inline temp(x, y, z, t) = 2.4 * cos(t * 2π / year + 50day) + 10
 nothing #hide
 
+architecture = GPU()
```
jagoosw, line 41:
```suggestion
architecture = CPU()
```

jagoosw (approved):
This has been such a pain before, thank you for fixing!

jagoosw:
I don't think the "Simulations Tips" is the correct place but is there somewhere in the Oceananigans repo that would be appropriate for me to add some notes about writing code for GPU? I think there are some non-obvious things I've picked up in the last few months in addition to this point about not passing `models` to kernel functions such as needing to write `adapt_structure` methods for new structs that might be useful to other people.

navidcy:
https://github.com/CliMA/Oceananigans.jl/wiki ?

jagoosw:
Yeah that probably makes sense, I'll add a dev tips section to the GPU page

navidcy:
You should have edit access for the wiki; do you?

jagoosw:
> You should have edit access for the wiki; do you?

Yeah I've got access

PR #110, opened 2023-06-23
Update Oceananigans/KernelAbstractions versions

jagoosw:


jagoosw:
With the new kernel launching, do I need to do something to get the threads to sync now that there's no `wait!(event)` going on? I can't see anything that looks like it tries to synchronise things in the Oceananigans repo

jagoosw:
This will be a breaking change due because of removing the interface to specify advection schemes for the drift velocity (since Oceananigans no longer supports in since https://github.com/CliMA/Oceananigans.jl/pull/3027

jagoosw:
Also, revisiting how I've implemented sediment integration because of some breaking changes from Oceananigans has reaffirmed to me that I still don't like how I've implemented it because it has to repeat so much from Oceananigans and uses non-public functions that break often

navidcy:
Is it a good idea to test some of the examples on a GPU since the tests don’t run on GPU as far as I recall?

jagoosw:
> Is it a good idea to test some of the examples on a GPU since the tests don’t run on GPU as far as I recall?

That's a good idea, I'll try running the examples on our GPU node.

I do intend to setup GPU testing like Oceananigans has at some point but it seems relatively challenging?

navidcy:
> I do intend to setup GPU testing like Oceananigans has at some point but it seems relatively challenging?

I wan't part of setting that up -- not sure how to set the tests to run on your own machine.

jagoosw:
The examples `column.jl`, `kelp.jl`, and `eady.jl` (in 16s!) all seem to work just by changing the architecture to `GPU` and I think between them they should cover everything except the NPZD model. I have also realised that the NPZD example just doesn't work, so I will fix.

navidcy:
Is there a reason the NPZD example is not included in the list of examples that are literated in the docs?

jagoosw:
I was just wondering this, I've just had a go at running it and I don't think I ever really finished it. Perhaps we should just get rid of it?

jagoosw:
This is a video of the first 10 days and I don't think it looks like I ever set it up to be physically interesting, or tuned the initial conditions to get the BGC to show anything.

https://github.com/OceanBioME/OceanBioME.jl/assets/26657828/fc71af55-c699-4f86-8fbb-509f2466411d

jagoosw:
Perhaps we could leave it out for now and sort it out elsewhere? Just I was going to submit the JOSS paper today and you have to put a version, so I wanted to merge this first because its got quite a lot of changes.

navidcy:
Yes, nuke the NPZD example. It'll be there in the git history and we can add it later!

OK! I'm happy to approve this PR since you confirm that you ran the examples on GPU and all is good.

jagoosw:
> Yes, nuke the NPZD example. It'll be there in the git history and we can add it later!
> 
> OK! I'm happy to approve this PR since you confirm that you ran the examples on GPU and all is good.

Okay great, thank you! I've also just realised that the quick start example also uses the NPZD model so it is getting run in the docs, so this example probably isn't very instructional anyway and is confusing

Manifest.toml
```diff
@@ -1,30 +1,59 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.3"
+julia_version = "1.9.1"
```
navidcy, line 3: seems that you made the manifest with v1.9.1
but since the tests run on v1.8 it's better to do use v1.8.5 to make the manifest, just in case there is a conflict or something

navidcy: I can fix this; shall I?

jagoosw: Yeah happy for you to fix, thanks

navidcy (approved):

navidcy:
Let's check the docs preview before we merge (just in case).

navidcy:
Docs look good! Merge?

navidcy:
[Note: Perhaps also deal with #112 before registering v0.2?]

johnryantaylor:
removing the NPZD example is fine since as Jago says an example of using this model is in the quickstart anyway (and the NZPD model is pretty straightforward)

jagoosw:
Great, yeah will merge #112 first. Thank you!

PR #112, opened 2023-06-23
Add GPU info in README

navidcy:
I tried it and indeed the example + the visualization works on a GPU machine with only the change `CPU()`->`GPU()` in the grid.


navidcy:
@jagoosw feel free to edit/modify/merge this as per you wish. I thought it would be nice to have something about GPUs somewhere.

jagoosw (approved):

PR #113, opened 2023-06-24
added advection back to examples

jagoosw:
Before the advection fix in Oceananigans you could turn off advection (i.e. set it to `nothing`) in column models because the drift velocities had their own advection schemes and it sped up run time. Now that all velocities are treated together we can't do this anymore so need an advection scheme to be on for particle sinking to work correctly.

(@syou83syou83 you may need to update this in your scripts if you update your OceanBioME version)

### DIFF
--- a/examples/column.jl
+++ b/examples/column.jl
@@ -49,8 +49,7 @@ model = NonhydrostaticModel(; grid,
                               biogeochemistry = LOBSTER(; grid,
                                                           surface_phytosynthetically_active_radiation = PAR⁰,
                                                           carbonates = true),
-                              boundary_conditions = (DIC = FieldBoundaryConditions(top = CO₂_flux), ),
-                              advection = nothing)
+                              boundary_conditions = (DIC = FieldBoundaryConditions(top = CO₂_flux), ))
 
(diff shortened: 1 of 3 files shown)


navidcy:
Do we need to bump a patch release?

jagoosw:
Ah yeah forgot that

navidcy:
Did the examples in the docs look weird and I missed it? sorry

navidcy:
To be honest I don't understand what the issue is. But if you have checked and you are sure then I can approve.

jagoosw:
> Did the examples in the docs look weird and I missed it? sorry

No dw, it only makes a very subtle difference in the docs (you can kind of see the detritus concentration is off, and the sinking flux line changes shape slightly) but I only noticed it when I ran a longer model testing something else.

navidcy (approved):

jagoosw:
I'm not sure how patch versioning works. For breaking changes, we update the minor version, Y in X.Y.Z, and for any other changes, we update Z and release a package version?

navidcy:
yeah, bug fixes/new feat/anything non-breaking you bump up the Z :)
you can wait for #114 before you release so that the docs stable include it

PR #120, opened 2023-07-04
Generalize  `TwoBandPhotosyntheticallyActiveRadiation` kernel to allow curvilinear grids

### LINKED ISSUE #119: Light attenuation model does not work on lat/lon grids
Light attenuation model does not work on lat/lon grids because `xnode` is defined differently for each. I imagine this may have been fixed recently in Oceananigans so will just be a case of updating the use of the `xnode` function.

jagoosw:
Not sure this is the most efficient way todo this, but since `xnode` is defiend differedntly for different grid types this is by far the easiest fix. The alternative would be to define `xnode(i, grid, lx, ly, lz)` for RectilinearGrids. Perhaps that solution would be better.

Closes #119

### DIFF
--- a/src/Boundaries/Sediments/simple_multi_G.jl
+++ b/src/Boundaries/Sediments/simple_multi_G.jl
@@ -90,7 +90,7 @@ function SimpleMultiG(grid;
                                            D = -1.3349, 
                                            E = 0.1826, 
                                            F = - 0.0143),
-                      depth = abs(znode(1, grid, Face())),
+                      depth = abs(znodes(grid, Face())[1]),
                       solid_dep_params::P4 = (A = 0.233, 
                                               B = 0.336, 
                                               C = 982, 
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,8 +1,8 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x = xnode(i, j, 1, grid, Center(), Center(), Center())
-    y = ynode(i, j, 1, grid, Center(), Center(), Center())
+    x = node(i, j, 1, grid, Center(), Center(), Center())
+    y = node(i, j, 1, grid, Center(), Center(), Center())
     
     PAR⁰ = surface_PAR(x, y, t)
 
@@ -89,7 +89,9 @@ Keyword Arguments
 
 - `grid`: grid for building the model on
 - `water_red_attenuation`, ..., `phytoplankton_chlorophyll_ratio`: parameter values
-- `surface_PAR`: function (or array in the future) for the photosynthetically available radiation at the surface, should be shape `f(x, y, t)`
+- `surface_PAR`: function (or array in the future) for the photosynthetically available radiation at the surface, 
+   which should be `f(x, y, t)` where `x` and `y` are the native coordinates (i.e. meters for rectilinear grids
+   and latitude/longitude as appropriate)
 """
 function TwoBandPhotosyntheticallyActiveRadiation(; grid, 
                                                     water_red_attenuation::FT = 0.225, # 1/m
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,7 +1,8 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x, y = @inbounds xnodes(grid, Center(), Center(), Center())[i], ynodes(grid, Center(), Center(), Center())[j]
+    x = xnode(i, j, 1, grid, Center(), Center(), Center())
+    y = ynode(i, j, 1, grid, Center(), Center(), Center())
     
     PAR⁰ = surface_PAR(x, y, t)
 
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,8 +1,7 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x = node(i, j, 1, grid, Center(), Center(), Center())
-    y = node(i, j, 1, grid, Center(), Center(), Center())
+    x, y, _ = node(i, j, 1, grid, Center(), Center(), Center())
     
     PAR⁰ = surface_PAR(x, y, t)
 
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,7 +1,7 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x, y = @inbounds xnodes(grid, Center(), Center(), Center())[i], ynodes(j, grid, Center(), Center(), Center())[j]
+    x, y = @inbounds xnodes(grid, Center(), Center(), Center())[i], ynodes(grid, Center(), Center(), Center())[j]
     
     PAR⁰ = surface_PAR(x, y, t)
 
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,7 +1,7 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x, y = xnode(i, grid, Center()), ynode(j, grid, Center())
+    x, y = @inbounds xnodes(grid, Center(), Center(), Center())[i], ynodes(j, grid, Center(), Center(), Center())[j]
     
     PAR⁰ = surface_PAR(x, y, t)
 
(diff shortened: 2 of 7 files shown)


jagoosw:
Probably should add some tests on Lat/lon grids

navidcy:
What do you mean by

> but since `xnode` is defined differently for different grid types this is by far the easiest fix

?

navidcy:
```Julia
xnode(i, j, k, grid, ℓx, ℓy, ℓz)
```

should work for _all_ grids. If the above is not true then we should fix it in Oceananigans.

Note that `xnode(s)` will return the zonal coordinate (e.g. in meters), not the grid's native x-coordinate which can be degrees for some grids.

src/Light/2band.jl
```diff
@@ -1,7 +1,7 @@
 @kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model) 
     i, j = @index(Global, NTuple)
 
-    x, y = xnode(i, grid, Center()), ynode(j, grid, Center())
+    x, y = @inbounds xnodes(grid, Center(), Center(), Center())[i], ynodes(grid, Center(), Center(), Center())[j]
```
navidcy, line 4: not sure if `k=1` is what we want here, but how about:
```suggestion
    x = xnode(i, j, 1, grid, Center(), Center(), Center())
    y = ynode(i, j, 1, grid, Center(), Center(), Center())
```

glwagner: Note that for general curvilinear grids, the nodes depend on both `i, j`. So you probably want to use @navidcy's suggestion.

All current grids (and for the forseeable future) are independent of `k` so it's ok to take `k=1`. This makes a "thin spherical shell" approximation for the ocean.

jagoosw:
Thank you both @navidcy @glwagner . I'm not sure exactly what I've got confused by, I think it was trying to do `xnode(i, j, grid...` rather than with `i, j, k`. I'll apply these suggestions and see if it works.

I was trying to run one of the ClimaOcean examples with a BGC model which initially wasn't working because the light just had `xnode(i, grid...`, and then I think I just fixed it wrong.

navidcy:
Note that ClimaOcean might need some updating!

jagoosw:
It does work now

> Note that ClimaOcean might need some updating!

I did have to remove some compat entries to install OceanBioME with it but it seems to be running fine with Oceananigans 0.83.0

src/Boundaries/Sediments/simple_multi_G.jl
```diff
@@ -90,7 +90,7 @@ function SimpleMultiG(grid;
                                            D = -1.3349, 
                                            E = 0.1826, 
                                            F = - 0.0143),
-                      depth = abs(znode(1, grid, Face())),
+                      depth = abs(znodes(grid, Face())[1]),
```
navidcy, line 93: `znodes` is user-facing and available as soon as you call `using Oceananigans` so I changed  it here.
this is only called once so it doesn't matter if it's not the most efficient way

src/Models/Individuals/SLatissima.jl
navidcy, line 1: shall we delete commented code? it's unclear why it's there..
```
    #nodes, weights, normfactor = @inbounds get_nearest_nodes(x, y, z, grid, (Center(), Center(), Center()))
    #@unroll for (n, weight) in enumerate(weights)
```

jagoosw: Oh yeah, not sure why that's still there

jagoosw:
I've just thought, it's probably more convenient for a user if the surface PAR is in the grid native coordinates, for example, if doing a global simulation data is going to be in lat/lon coordinates. Using `Oceananigans.Grids.node` would be the appropriate thing to use here then right?

navidcy:
```Julia
node(i, j, k, grid, ℓx, ℓy, ℓz)
```
will return a 1-, 2-, or 3- tuple with the native coordinates of the grid. (1-, 2-, or 3-tuple depending if any of the coordinates are Flat).

jagoosw:
> ```julia
> node(i, j, k, grid, ℓx, ℓy, ℓz)
> ```
> 
> will return a 1-, 2-, or 3- tuple with the native coordinates of the grid. (1-, 2-, or 3-tuple depending if any of the coordinates are Flat).

Oh yeah I mean to fix this

navidcy:
hm... I think what you did won't work... because you seem to assume the `node` will return a 3-tuple but it won't necessarily do that

navidcy:
So I'm not sure what you are trying to do so I can't be very specific.
But how about this, is this helpful?
```Julia
locs = node(i, j, k, grid, Center(), Center(), Center())

value = surface_PAR(loss...)
```

This way you don't unfold the tuple but just pass it to the `surface_PAR` as a tuple and use `...` to unpack it then?

jagoosw:
I just checked and it seems to work on flat dimensions always returning a 3-tuple, as long as the locations are all not `nothing` because that's what the different length results are dispatched on:

https://github.com/CliMA/Oceananigans.jl/blob/92791a962c9096746301cdb888a3050e32c4a58b/src/Grids/latitude_longitude_grid.jl#L624-L634

jagoosw:
> So I'm not sure what you are trying to do so I can't be very specific. But how about this, is this helpful?
> 
> ```julia
> locs = node(i, j, k, grid, Center(), Center(), Center())
> 
> value = surface_PAR(loss...)
> ```
> 
> This way you don't unfold the tuple but just pass it to the `surface_PAR` as a tuple and use `...` to unpack it then?

I guess this would work although we currently have `surface_PAR` as a function of x and y only, but if this doesn't work we can change that to include z

navidcy:
> I just checked and it seems to work on flat dimensions always returning a 3-tuple, as long as the locations are all not `nothing` because that's what the different length results are dispatched on:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/92791a962c9096746301cdb888a3050e32c4a58b/src/Grids/latitude_longitude_grid.jl#L624-L634

Hm... But this, e.g.,
https://github.com/CliMA/Oceananigans.jl/blob/92791a962c9096746301cdb888a3050e32c4a58b/src/Grids/latitude_longitude_grid.jl#L629
is a 2-tuple.

navidcy:
> > So I'm not sure what you are trying to do so I can't be very specific. But how about this, is this helpful?
> > ```julia
> > locs = node(i, j, k, grid, Center(), Center(), Center())
> > 
> > value = surface_PAR(loss...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way you don't unfold the tuple but just pass it to the `surface_PAR` as a tuple and use `...` to unpack it then?
> 
> I guess this would work although we currently have `surface_PAR` as a function of x and y only, but if this doesn't work we can change that to include z

Sorry, ignore my suggestion then.

Perhaps I just don't know what you are trying to do and I was assuming things.

Let's restart: Yes, to get the coordinates of the grid at `i, j, k` then `node(i,j, k, grid,...)` will do the job. :)

jagoosw:
> > I just checked and it seems to work on flat dimensions always returning a 3-tuple, as long as the locations are all not `nothing` because that's what the different length results are dispatched on:
> > https://github.com/CliMA/Oceananigans.jl/blob/92791a962c9096746301cdb888a3050e32c4a58b/src/Grids/latitude_longitude_grid.jl#L624-L634
> 
> Hm... But this, e.g., https://github.com/CliMA/Oceananigans.jl/blob/92791a962c9096746301cdb888a3050e32c4a58b/src/Grids/latitude_longitude_grid.jl#L629 is a 2-tuple.

Since we're always giving the locs as Center/Center/Center won't it always use the first method though? I guess it uses the others if you were to e.g. do `locations(velocity field)` which would return a nothing location on the flat dimension, and then use the function?

navidcy:
> Since we're always giving the locs as Center/Center/Center won't it always use the first method though? I guess it uses the others if you were to e.g. do `locations(velocity field)` which would return a nothing location on the flat dimension, and then use the function?

True! My bad...

navidcy:
should we add a test?

jagoosw:
I've changed the light test to test on different grids and models now, I don't think we need the other tests todo this since the others just test the properties of the bgc model components and their ability to work in different grids etc. should be being tested in the Oceananigans tests right?

I think maybe I should add a test that the particle tendencies and sediment models work across all model types. Although again their ability to function should be caught by the Oceananigans tests now. Perhaps I should test their combinations since that's an issue I found recently.

jagoosw:
I forgot I hadn't pushed the sediment model fixes. I've just made this PR  #121 which has them so I think we should merge these tests and fixes, then look at adding more sediment model tests in that PR.

test/test_light.jl
```diff
@@ -38,5 +40,22 @@ grid = RectilinearGrid(size=(1,1,2), extent=(1,1,2))
 
     results_PAR = convert(Array, model.biogeochemistry.light_attenuation_model.field)[1, 1, 1:2]
 
-    @test all(results_PAR .≈ reverse(expected_PAR))
+    return all(results_PAR .≈ reverse(expected_PAR))
+end
+
+archs = (CPU(), )
+
+@testset "Light attenuaiton model" begin
```
navidcy, line 48:
```suggestion
@testset "Light attenuation model" begin 
```

test/test_light.jl
```diff
@@ -1,22 +1,24 @@
...
+                            surface_phytosynthetically_active_radiation = (x, y, t) -> 100.0)
 
-    model = NonhydrostaticModel(; grid, 
-                                  biogeochemistry = LOBSTER(; grid,
-                                                              light_attenuation_model = TwoBandPhotosyntheticallyActiveRadiation(; grid),
-                                                              surface_phytosynthetically_active_radiation = (x, y, t) -> 100.0))
-    Pᵢ(x,y,z) = 2.5 + z
+    model = model_type(; grid, 
+                         biogeochemistry,
+                         tracers = unique((required_biogeochemical_tracers(biogeochemistry)..., :T, :S))) # because hydrostatic free surface will request T and S and some BGC models will too
```
navidcy, line 14:
```suggestion
                         tracers = unique((required_biogeochemical_tracers(biogeochemistry)..., :T, :S))) # because hydrostatic free surface model will request T and S and some BGC models will too
```

navidcy (approved):
lgtm

PR #121, opened 2023-07-13
Sediment model fixes + simpler model

jagoosw:
I've finally fixed sediment models so they work properly and conserve nitrogen!

Also fixed a few bugs in the sediment model time-stepping and adds a simple instant remineralization model (as described in PISCESv2 paper).

Hopefully after #120 is merged the odd commit history will be fixed.

PR #122, opened 2023-07-15
Update Oceananigans dependency + run CI and Docs on julia v1.9

navidcy:
@jagoosw, is updating to Oceananigans v0.84.1 a good idea?


jagoosw:
Yeah seems like a good idea. I'm thinking for merging all the current PRs, if I do #124 and #123 which are non-breaking and quite simple, then #121, then fix the conflicts here and bump the version here?

jagoosw (approved):
lgtm

navidcy:
> Yeah seems like a good idea. I'm thinking for merging all the current PRs, if I do #124 and #123 which are non-breaking and quite simple, then #121, then fix the conflicts here and bump the version here?

is it a breaking change?

navidcy:
OK, I think we are hitting this: https://github.com/CliMA/Oceananigans.jl/issues/3184 and tests fail on v1.8.

navidcy:
So on second thought, post v0.84.1 Oceananigans is _only_ tested on Julia v1.9. So perhaps it's safer to drop v1.8?

jagoosw:
Yeah I'm happy to drop 1.8 support

jagoosw:
> > Yeah seems like a good idea. I'm thinking for merging all the current PRs, if I do #124 and #123 which are non-breaking and quite simple, then #121, then fix the conflicts here and bump the version here?
> 
> is it a breaking change?

The sediment stuff defiantly breaks the sediment API, I guess this isn't breaking though so could do this first then release on #121

navidcy:
Well it would be breaking if we drop v1.8! :)

navidcy:
I don't know why there is the 

> Julia 1.8 - ubuntu-latest - x64 - pull_request Expected — Waiting for status to be reported

That shouldn't be there. I removed those tests... I'll merge when everything else passes.

jagoosw:
I think its because the branch protection rule is specific to the 1.8 test, I've changed it to the 1.9 workflow now so I'm not sure why its still there. So yeah bypass the rule and I think it should work for the next PR

PR #123, opened 2023-07-15
Remove DiffEqBase from dependencies

navidcy:
I don't see this being used anywhere. Am I wrong?


jagoosw (approved):
Ah yeah I think this is not used anymore, thanks

PR #127, opened 2023-08-04
(0.5.0) Updated sediment to work with all grids and model types

jagoosw:
This PR makes sediment models work with immersed boundary grids and improves sediment testing on all grid and model types.

This breaks the sediment, nitrogen/carbon flux, and sinking flux APIs.

### DIFF
--- a/examples/eady.jl
+++ b/examples/eady.jl
@@ -81,7 +81,16 @@ vᵢ(x, y, z) = Ũ * Ξ(z)
 
 set!(model, u=uᵢ, v=vᵢ, P = 0.03, Z = 0.03, NO₃ = 4.0, NH₄ = 0.05, DIC = 2200.0, Alk = 2409.0)
 
-simulation = Simulation(model, Δt = 15minutes, stop_time = 10days)
+# ## Setup the simulation
+# Choose an appropriate initial timestep for this resolution and set up the simulation
+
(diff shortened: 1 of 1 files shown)


jagoosw:
Think this is probably the most straightforward way to set this up, and in this case the bottom cell only has to be calculated once

jagoosw:
I think in the same vein it may be worth modifying PAR integration to only integrate down to the bottom cell since a lot of the globe doesn't reach the bottom of the underlying grid (and at the moment I think we integrate it for the land area too)

jagoosw:
Since it's just the codecov failing (and I'm not bothered by the lines in the NPZD model it says aren't being tested as it's just a warning and function no one will probably ever use) I'm happy for this PR to be merged now

jagoosw:
The example was failing because I'd changed the sediment models to only use flat fields, but we can't output these still due to https://github.com/CliMA/Oceananigans.jl/issues/2770 which I've not fixed yet

jagoosw:
Maybe sediment models should actually live on different grids. It might be more flexible in the future if someone wanted to implement a not-flat model, and would probably solve the output issue.
It probably also wouldn't be that hard to implement.

jagoosw:
Nonhydrostatic + RK3 is not running getting tested but it shoudl

PR #130, opened 2023-08-15
Update README.md

johnryantaylor:
fixed a few typos in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ end
 ```
 </details>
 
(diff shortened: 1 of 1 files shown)


navidcy (approved):
Good catch

jagoosw:
We need to add the new GIF back in too so should do that here

navidcy:
> We need to add the new GIF back in too so should do that here

@jagoosw can you do that?

PR #132, opened 2023-08-16
Update NPZ.md (resize the image)

iuryt:
Not sure this is the best practice, but I noticed the NPZ image was too large and suggested to use `` tag to resize the image. I used a size that the resulting text from the image is around the same size as the docs.


jagoosw:
Thank you. I'm not sure if it works in this version of markdown but you can sometimes do:

```
![alt text](url){width = "700px"}
```

jagoosw (approved):

PR #133, opened 2023-08-25
Update model_implementation.md

johnryantaylor:
Made some edits to the model implementation page to try to clarify a few aspects (and a few small corrections)

### DIFF
--- a/docs/src/model_implementation.md
+++ b/docs/src/model_implementation.md
@@ -15,7 +15,7 @@ The nature of multiple dispatch in Julia means that we define new BGC models as 
 
 For this example we are going to implement the simple Nutrient-Phytoplankton model similar to that used in [Chen2015](@citep), although we neglect the nutrient in/outflow terms since they may be added as [boundary conditions](https://clima.github.io/OceananigansDocumentation/stable/model_setup/boundary_conditions/), and modified to conserve nitrogen.
 
-The first step is to import the abstract type from OceanBioME, some units from Oceananigans (for ease of parameter definition), and [`import`](https://stackoverflow.com/questions/27086159/what-is-the-difference-between-using-and-import-in-julia-when-building-a-mod) some functions from Oceananigans which we will add methods to:
+The first step is to import the abstract type from OceanBioME, some units from Oceananigans (for ease of parameter definition), and [`import`](https://stackoverflow.com/questions/27086159/what-is-the-difference-between-using-and-import-in-julia-when-building-a-mod) some functions from Oceananigans in order to add methods to:
 
 ```@example implementing
 using OceanBioME: ContinuousFormBiogeochemistry
@@ -48,7 +48,7 @@ We then define our `struct` with the model parameters, as well as slots for the 
 end
 ```
 
-Here, we use descriptive names for the parameters. Below, each of these will correspond to a symbol (or letter) which is more convenient mathematically and when defining the BGC model functions. In the above code we used `@kwdef` to set default values for the models so that we don't have to set all of these parameters each time we use the model. The default parameter values can optionally be over-ridden by the user when they run the mmodel. We have also included a `sinking_velocity` field in the parameter set to demonstrate how we can get tracers (e.g. detritus) to sink. We also need to define some functions so that OceanBioME and Oceananigans know what tracers and auxiliary fields (e.g. light intensity) we will use:
+Here, we use descriptive names for the parameters. Below, each of these parameters correspond to a symbol (or letter) which is more convenient mathematically and when defining the BGC model functions. In the above code we used `@kwdef` to set default values for the models so that we don't have to set all of these parameters each time we use the model. The default parameter values can optionally be over-ridden by the user when running the model. We have also included a `sinking_velocity` field in the parameter set to demonstrate how we can get tracers (e.g. detritus) to sink. We also need to define some functions so that OceanBioME and Oceananigans know what tracers and auxiliary fields (e.g. light intensity) we will use:
 
 ```@example implementing
 required_biogeochemical_tracers(::NutrientPhytoplankton) = (:N, :P, :T)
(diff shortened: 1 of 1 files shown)


docs/src/model_implementation.md
```diff
@@ -15,7 +15,7 @@ The nature of multiple dispatch in Julia means that we define new BGC models as
 
 For this example we are going to implement the simple Nutrient-Phytoplankton model similar to that used in [Chen2015](@citep), although we neglect the nutrient in/outflow terms since they may be added as [boundary conditions](https://clima.github.io/OceananigansDocumentation/stable/model_setup/boundary_conditions/), and modified to conserve nitrogen.
 
-The first step we need is to import the abstract type from OceanBioME, some units from Oceananigans (for ease of parameter definition), and [`import`](https://stackoverflow.com/questions/27086159/what-is-the-difference-between-using-and-import-in-julia-when-building-a-mod) some functions from Oceananigans which we will add methods to:
+The first step is to import the abstract type from OceanBioME, some units from Oceananigans (for ease of parameter definition), and [`import`](https://stackoverflow.com/questions/27086159/what-is-the-difference-between-using-and-import-in-julia-when-building-a-mod) some functions from Oceananigans which we will add methods to:
```
navidcy, line 18:
```suggestion
The first step is to import the abstract type from OceanBioME, some units from Oceananigans (for ease of parameter definition), and [`import`](https://stackoverflow.com/questions/27086159/what-is-the-difference-between-using-and-import-in-julia-when-building-a-mod) some functions from Oceananigans in order to add methods to:
```

docs/src/model_implementation.md
```diff
@@ -48,7 +48,7 @@ We then define our `struct` with the model parameters, as well as slots for the
 end
 ```
 
-In the above we used `@kwdef` to set default values for the models so that we don't have to set these up each time we use the model. We have also included a `sinking_velocity` field to later demonstrate how we can get tracers to sink. We also need to define some functions so that OceanBioME and Oceananigans know what tracers and auxiliary fields (e.g. light intensity) we need setting up:
+Here, we use descriptive names for the parameters. Below, each of these will correspond to a symbol (or letter) which is more convenient mathematically and when defining the BGC model functions. In the above code we used `@kwdef` to set default values for the models so that we don't have to set all of these parameters each time we use the model. The default parameter values can optionally be over-ridden by the user when they run the mmodel. We have also included a `sinking_velocity` field in the parameter set to demonstrate how we can get tracers (e.g. detritus) to sink. We also need to define some functions so that OceanBioME and Oceananigans know what tracers and auxiliary fields (e.g. light intensity) we will use:
```
navidcy, line 51:
```suggestion
Here, we use descriptive names for the parameters. Below, each of these parameters correspond to a symbol (or letter) which is more convenient mathematically and when defining the BGC model functions. In the above code we used `@kwdef` to set default values for the models so that we don't have to set all of these parameters each time we use the model. The default parameter values can optionally be over-ridden by the user when running the model. We have also included a `sinking_velocity` field in the parameter set to demonstrate how we can get tracers (e.g. detritus) to sink. We also need to define some functions so that OceanBioME and Oceananigans know what tracers and auxiliary fields (e.g. light intensity) we will use:
```

docs/src/model_implementation.md
```diff
@@ -58,23 +58,22 @@ required_biogeochemical_auxiliary_fields(::NutrientPhytoplankton) = (:PAR, )
 biogeochemical_auxiliary_fields(bgc::NutrientPhytoplankton) = biogeochemical_auxiliary_fields(bgc.light_attenuation_model)
 ```
 
-Next, we need to define the methods that specify how the nutrient, ``N``, and phytoplankton ``P`` evolve. We want the phytoplankton to evolve at the rate given by:
-
+Next, we define the functions that specify how the phytoplankton ``P`` evolve. In the absence of advection and diffusion (which will be handled by Oceananigans), we want the phytoplankton to evolve at the rate given by:
```
navidcy, line 61:
```suggestion
Next, we define the functions that specify how the phytoplankton ``P`` evolve. In the absence of advection and diffusion (both of which are handled by Oceananigans), we want the phytoplankton to evolve at the rate given by:
```

docs/src/model_implementation.md
```diff
@@ -58,23 +58,22 @@ required_biogeochemical_auxiliary_fields(::NutrientPhytoplankton) = (:PAR, )
...
 ```math
 \frac{\partial P}{\partial t} = \mu g(T) f(N) h(PAR) P - mP - bP^2,
 ```
 
-where ``\mu`` is `base_growth_rate`, ``m`` is `mortality_rate`, and ``b`` is `crowding`. The functions ``g``, ``f``, and ``h`` are given by:
+where $\mu$ corresponds to the parameter `base_growth_rate`, ``m`` corresponds to the parameter `mortality_rate`, and ``b`` corresponds to the parameter `crowding_mortality_rate`. Here, the functions ``g``, ``f``, and ``h`` are defined by:
```
navidcy, line 66:
```suggestion
where ``\mu`` is the parameter `base_growth_rate`, ``m`` is the parameter `mortality_rate`, and ``b`` is the parameter `crowding_mortality_rate`. Here, the functions ``g``, ``f``, and ``h`` are:
```

docs/src/model_implementation.md
```diff
@@ -110,13 +109,18 @@ end
...
+```math
+\frac{\partial N}{\partial t} = - \frac{\partial P}{\partial t}
+```
+Hence, we will define the nutrient forcing using as the negative of the phytoplankton forcing
 ```@example implementing
 @inline (bgc::NutrientPhytoplankton)(::Val{:N}, args...) = -bgc(Val(:P), args...)
 ```
 
-Finally, we need to define some functions to make sure that the sub-models (light attenuation) get updated in the right place
+Finally, we need to define some functions to allow us to update the time-dependent parameters (in this case the PAR and temperature, T):
```
navidcy, line 123:
```suggestion
Finally, we need to define some functions to allow us to update the time-dependent parameters (in this case the PAR and temperature, ``T``):
```

navidcy (approved):
lgtm!

jagoosw (approved):

docs/src/model_implementation.md
```diff
@@ -58,23 +58,22 @@ required_biogeochemical_auxiliary_fields(::NutrientPhytoplankton) = (:PAR, )
...
+where $\mu$ corresponds to the parameter `base_growth_rate`, ``m`` corresponds to the parameter `mortality_rate`, and ``b`` corresponds to the parameter `crowding_mortality_rate`. Here, the functions ``g``, ``f``, and ``h`` are defined by:
 ```math
 \begin{align}
 g(T) &= c_1\exp\left(-c_2|T - T_{opt}|\right),\\
 f(N) &= \frac{N}{k_N + N},\\
 h(PAR) &= \frac{PAR}{k_P + PAR},
 \end{align}
 ```
-where ``c_1`` is `temperature_coefficient`,  ``c_2`` is `temperature_exponent`, ``T_{opt}`` is `optimal_temperature`, ``k_N`` is `nutrient_half_saturation`, and ``k_P`` is `light_half_saturation`. Since this is a simple two variable model and the total concentration is conserved, ``\frac{\partial N}{\partial t} = - \frac{\partial P}{\partial t}``.
+where $c\_1$ corresponds to `temperature_coefficient`,  $c\_2$ corresponds to `temperature_exponent`, ``T_{opt}`` corresponds to `optimal_temperature`, ``k_N`` corresponds to `nutrient_half_saturation`, and ``k_P`` corresponds to `light_half_saturation`.
```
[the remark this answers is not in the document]
johnryantaylor: Does it still render symbolically without the dollar signs?  I made this change because it appeared as \mu and c_1, etc. in the preview of the .md file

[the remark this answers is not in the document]
navidcy: Yeap!
See https://oceanbiome.github.io/OceanBioME.jl/previews/PR133/model_implementation/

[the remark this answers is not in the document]
navidcy: (Github’s preview of md files might not work, but Documenter renders this all right!)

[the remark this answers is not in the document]
johnryantaylor: Oh, ok, thanks for clarifying.  It was on github where it looked off, but you're right that it is rendering correctly on documenter.

PR #134, opened 2023-08-27
(0.6.0) Refactor models for more abstraction (and bumped compats)

jagoosw:
I've been trying to work out how to better separate the sub-models (bgc, sediment, light attenuation, particles) without having to repeat lots of methods for each bgc model and think this abstraction achieves that.

I've not tried all of the tests yet but at least some are working locally.

Todo:

- [x] Get all tests working
- [x] Add abstract model docstring
- [x] Update model docstrings
- [x] Update documentation
- [x] Get doctests to work
- [x] Introduce a `redfield` function to get the redfield ratio of a tracer
- [x] Update LOBSTER tests to use `redfield` and `conserved_tracers` as an example of how they work (and to tidy up the mess)

I've attempted to introduce this `redfield` function as I think it should simplify some code, and should prevent scripts from breaking as structs change etc in the same vaine as Oceanaingans' `nodes`. For example in the `LOBSTER` tests we should be able to reduce some of the complexity in the carbon sum function (but not done this yet).

I don't really want to update the docs (think it would just be the model implementation page) until my other PR is merged otherwise I'll have todo it twice.

This PR does not strictly cause breaking changes (I think), but since we want other people to implement models I think we should consider it breaking as it would cause other model implementations to stop working. Also since we now export `Biogeochemistry` changes like this in the future probably will be breaking.

This update should also mean that any Oceananigans biogeochemical model can easily be coupled with the other things we provide like sediment models, where as previously a user would have to redefne their model with our infrastructure.

> (NB: this branch has an inappropriate name because I started off doing something else and forgot to change it before I started this PR)


jagoosw:
I'm going to change the target test coverage for codecov (if I can work out how) since I think we currently have adequate coverage but it always looks like tests are failing because it doesn't meet the default target.

jagoosw:
The data forced example appears to be broken but otherwise done here

jagoosw:
@johnryantaylor I've improved how we prevent negative tracers in this PR as discussed in https://github.com/OceanBioME/OceanBioME.jl/issues/135 and it seems a lot more reliable now (and I think the setup is cleaner)

johnryantaylor:
fantastic, thanks!

docs/src/model_components/biogeochemical/index.md
```diff
@@ -5,7 +5,7 @@ Biogeochemical (BGC) models can be used within the [Oceananigans biogeochemistry
 For details of the BGC models currently implemented please see the following pages.
 
 ## Oceananigans setup
-At the simplest level all that is required to setup a BGC model is to pass it to the Oceananigans model setup:
+At the simplest level all that is required to setup an existint OceanBioME BGC model is to pass it to the Oceananigans model setup:
```
glwagner:
```suggestion
At the simplest level all that is required to setup an existing OceanBioME BGC model is to pass it to the Oceananigans model setup:
```

jagoosw: I think this is fixed in the current version right? https://github.com/OceanBioME/OceanBioME.jl/blob/41be828536b5f670a003eaeabffa819c77620223/docs/src/model_components/biogeochemical/index.md?plain=1#L8

glwagner: If you say so, I believe it!

Project.toml
```diff
@@ -12,12 +12,12 @@ Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 
 [compat]
-Adapt = "3.4"
-JLD2 = "0.4.31"
+Adapt = "3"
+JLD2 = "^0.4"
```
navidcy, line 16:
```suggestion
JLD2 = "0.4"
```

(it's the same, right?)

navidcy (approved):
only minor comments
code looks cleaner! I didn't check the actual implementation but I trust that since the tests pass it was done correctly

@jagoosw if there is any particular part of the code that changed that you'd like me to have a deep look let me know

Project.toml
jagoosw: Oh yeah think so!

docs/src/quick_start.md
```diff
@@ -30,13 +30,17 @@ simulation.output_writers[:profiles] = JLD2OutputWriter(model, model.tracers,
 run!(simulation)
 ```
 
-This isn't quite as simple as it could be as it records the output so that we can visualize it:
+This isn't quite as simple as it could be as it records the output so that we can visualize it, first check the required packages are installed:
```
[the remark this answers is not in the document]
navidcy: I'm sorry but here we set up a simulation, run it, save output and plot it in 30 lines??
Can you do this thing with any other code?

I don't think we should be apologetic.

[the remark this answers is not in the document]
jagoosw: Agreed!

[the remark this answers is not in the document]
navidcy: I mean I won't put my foot down on this. But I found it a bit strange..

jagoosw:
Happy for me to merge?

johnryantaylor (approved):
lgtm, thanks!

PR #136, opened 2023-08-28
Update eady.jl

johnryantaylor:
reduced initial timestep to 1 minute in Eady.jl to be more conservative and avoid CFL violation on finer grids

### DIFF
--- a/examples/eady.jl
+++ b/examples/eady.jl
@@ -81,7 +81,16 @@ vᵢ(x, y, z) = Ũ * Ξ(z)
 
 set!(model, u=uᵢ, v=vᵢ, P = 0.03, Z = 0.03, NO₃ = 4.0, NH₄ = 0.05, DIC = 2200.0, Alk = 2409.0)
 
-simulation = Simulation(model, Δt = 1minutes, stop_time = 10days)
+# ## Setup the simulation
+# Choose an appropriate initial timestep for this resolution and set up the simulation
+
+Δx = minimum_xspacing(grid, Center(), Center(), Center())
+Δy = minimum_yspacing(grid, Center(), Center(), Center())
+Δz = minimum_zspacing(grid, Center(), Center(), Center())
+
+Δt₀ = 0.75 * min(Δx, Δy, Δz) / maximum(V_field)
+
+simulation = Simulation(model, Δt = Δt₀, stop_time = 10days)
 
 # Adapt the time step while keeping the CFL number fixed.
 wizard = TimeStepWizard(cfl = 0.75, diffusive_cfl = 0.75, max_Δt = 30minutes)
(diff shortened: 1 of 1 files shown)


jagoosw:
Could we do:
```julia
v₀ = V(0, 0, 0, 0, background_state_parameters)

Δx = minimum_xspacing(grid, Center(), Center(), Center())
Δy = minimum_yspacing(grid, Center(), Center(), Center())
Δz = minimum_zspacing(grid, Center(), Center(), Center())

Δt₀ = 0.75 * min(Δx, Δy, Δz) / v₀

simulation = Simulation(model, Δt = Δt₀, stop_time = 10days)
```
so that it is setup right whatever grid size someone uses?

johnryantaylor:
Good idea - this might work

examples/eady.jl
```diff
@@ -81,7 +81,7 @@ vᵢ(x, y, z) = Ũ * Ξ(z)
 
 set!(model, u=uᵢ, v=vᵢ, P = 0.03, Z = 0.03, NO₃ = 4.0, NH₄ = 0.05, DIC = 2200.0, Alk = 2409.0)
 
-simulation = Simulation(model, Δt = 15minutes, stop_time = 10days)
+simulation = Simulation(model, Δt = 1minutes, stop_time = 10days)
```
jagoosw, line 84:
```suggestion
v₀ = V(0, 0, 0, 0, background_state_parameters)

Δx = minimum_xspacing(grid, Center(), Center(), Center())
Δy = minimum_yspacing(grid, Center(), Center(), Center())
Δz = minimum_zspacing(grid, Center(), Center(), Center())

Δt₀ = 0.75 * min(Δx, Δy, Δz) / v₀

simulation = Simulation(model, Δt = Δt₀, stop_time = 10days)
```

jagoosw (approved):

examples/eady.jl
navidcy: Is the maximum background state V at the bottom?

navidcy: something along `v_max =  maximum(V_field)` might be better?

jagoosw: It's got a $H + z$ term where H is a positive constant, but I guess it would be more general to take the maximum of the background field in case a user changes it.

navidcy: Either way I find that if the simulation errors because of CFL it should be OK that the user should need to figure it out. But if we can failsafe it a bit then it's better. If it's way too complicated to failsafe it then another option is when we prescribe the tilmestep to add a comment saying that "this choice works with resolution ..." or something giving a hint to the users that if they decide to run it at 10x resolution they might have to decrease the tilmestep.

PR #138, opened 2023-09-08
Make sediments work on GPU again

jagoosw:
We could do with some automated GPU testing


navidcy:
@jagoosw, does it make sense to wait for #139 before this? That way we'll have CI on GPU...

jagoosw:
@navidcy I'm not sure that #139 is going to work because the availability of GPU nodes for the tests to run on is so low it's taking a really long time for jobs to start. John and I have talked about a solution to this but I don't think it's going to be sorted in the short term. 

For now I have tested these changes manually on a GPU.

navidcy:
OK, so all tests pass on GPU in this PR? Good to know! Then I'll have a look at this soon. (Possibly after OSM abstract deadline..)

jagoosw:
> OK, so all tests pass on GPU in this PR? Good to know! Then I'll have a look at this soon. (Possibly after OSM abstract deadline..)

I haven't done them all but tested the sediment, I will let you know when I've run all of the tests.

jagoosw:
Okay a lot of the stuff is not working on GPU now, we need this https://github.com/CliMA/Oceananigans.jl/pull/3262 at least to make the negativity protection work again I think.

jagoosw:
Okay so now I'm actually getting around to testing everything on GPU I've discovered some more issues. The first of these is that scaling negative tracers can't pass tuples of symbols to kernels as discussed here https://github.com/CliMA/Oceananigans.jl/pull/3262, this is now solved.

The more difficult issue is that models are now often too large for the GPU parameter space (see https://github.com/JuliaGPU/CUDA.jl/issues/2080), I think this can be solved by only passing the relevant part of the model to kernels. 

My idea for this is to change adapt to return `adapt(to, biogeochemistry.underlying_biogeochemistry)` rather than the whole of `biogeochemistry` since all of the core Oceananigans uses of biogeochemistry in kernels only needs what is packaged in underlying biogeochemistry given that we discard the rest of it there anyway:
https://github.com/OceanBioME/OceanBioME.jl/blob/ac8419a1ce5a06a82be31588583636b212c09598/src/OceanBioME.jl#L120-L121
We can then ensure that in all other instances we only pass the relevant parts, e.g. here change from passing `bgc` to passing `bgc.underlying_biogeochemistry`:
https://github.com/OceanBioME/OceanBioME.jl/blob/ac8419a1ce5a06a82be31588583636b212c09598/src/Boundaries/Sediments/Sediments.jl#L38-L40

I think this would solve most of our issues as e.g. the light attenuation model seems to always be a lot of parameter space but only actually be used in update state. 

This is similar to how Oceananigans never passed models to kernels and hence doesn't have an `adapt_structure` method for it.

jagoosw:
> My idea for this is to change adapt to return `adapt(to, biogeochemistry.underlying_biogeochemistry)` rather than the whole of `biogeochemistry` since all of the core Oceananigans uses of biogeochemistry in kernels only needs what is packaged in underlying biogeochemistry given that we discard the rest of it there anyway:

This looks like it's working!

jagoosw:
I think I have now fixed all of the GPU bugs except:
- `column_diffusion_timescale` does not work on GPU and would need rethinking to do so.
- `SimpleMultiG` sediment kernels will now compile on GPU, but get the PTX parameter size error. I have marked the test for this one as broken and we can attempt to fix it if someone needs. These problems may go away at some point in the future: https://github.com/JuliaGPU/CUDA.jl/issues/2080

While not ideal I'm happy that all of the features people are currently using are tested and work on GPU. I will open issues for these two problems so that if people are trying to work out how to use them they have some informaiton.

jagoosw:
Closes #143

jagoosw:
Box models are broken again https://oceanbiome.github.io/OceanBioME.jl/previews/PR138/generated/box/

but apparently not all box models https://oceanbiome.github.io/OceanBioME.jl/previews/PR138/model_implementation/

jagoosw:
Tests all now work on GPU with the exception of Simple Multi G sediment with LOBSTER on Lat/lon and immersed boundary grids

test/test_sediments.jl
```diff
@@ -59,13 +66,15 @@ total_nitrogen(::LOBSTER, model) = sum(Array(interior(model.tracers.NO₃))) +
...
                                    sum(Array(interior(model.tracers.sPOM))) +
                                    sum(Array(interior(model.tracers.bPOM)))
+                                   
 total_nitrogen(::lobster_variable_redfield, model) = sum(Array(interior(model.tracers.NO₃))) +
                                                      sum(Array(interior(model.tracers.NH₄))) +
                                                      sum(Array(interior(model.tracers.P))) +
                                                      sum(Array(interior(model.tracers.Z))) +
                                                      sum(Array(interior(model.tracers.DON))) +
                                                      sum(Array(interior(model.tracers.sPON))) +
                                                      sum(Array(interior(model.tracers.bPON)))
```
glwagner: `sum(model.tracers.NO₃)` should work. It's preferred because it will omit immersed boundary regions, whereas writing `Array(interior(field))` will not.

glwagner: `sum(interior(field))` also works on the GPU and avoids the slow process of transferring data to the CPU.

navidcy: https://github.com/OceanBioME/OceanBioME.jl/blob/ac8419a1ce5a06a82be31588583636b212c09598/src/Models/AdvectedPopulations/LOBSTER/LOBSTER.jl#L453-L456

it's a collection of types; is `LobsterVariableRedfield` the suggested convention?

glwagner: Makes sense! Also Oceananigans style uses "english" semantics, so I think the type should be pronounced "VariableRedfieldLobster" as in "the variable-redfield lobster model"

jagoosw:
After propagating the changes from https://github.com/JuliaGPU/CUDA.jl/issues/2080 all of the tests now pass on GPU!

jagoosw:
> Btw I'm trying todo GPU tests here: https://colab.research.google.com/drive/1V4Kj2iTtxjsU44c5CkAmJjjQKbqR3tYd?usp=sharing

Update on this, AWS Sagemake has much higher GPU availability and resets your daily usage (4hours) each time you restart it + you can just run a terminal in it

jagoosw:
All tests now pass on GPU

jagoosw:
I'm fairly sure that the test coverage isn't actually reducing in this PR and its just fallbacks + CodeCov not dealing with Kernels very well. 

So I think this PR is ready now

johnryantaylor (approved):
lgtm, thanks!

PR #148, opened 2023-10-05
Add citation info in `README` + Docs

navidcy:


navidcy:
Perhaps we should add something in the Docs as well?

jagoosw:
I guess so, probably just the exact same on the home page?

jagoosw:
I think there is a mention about citing it on that page that can just be updated actually

navidcy:
Yes, true. I’ll do that ;)

jagoosw (approved):
lgtm, thanks!

PR #153, opened 2023-10-24
Accidentally deleted Oceananigans compat

jagoosw:


navidcy (approved):

navidcy:
I thought that the compat change was on purpose, in the sense that code didn’t work now with pre-v0.89.2 versions of Oceananigans. Wasn’t it?

PR #156, opened 2023-11-07
Fixed some `show` methods

jagoosw:
When I went to update Oceananigans I found these changes I made so thought we might as well merge them first too.

PR #155, opened 2023-11-04
(0.9.0) Fix typos in code, docs, and docstrings

pitmonticone:
This PR is a breaking change as it corrects spellings in the API. Changes in parameter names occur in:
- `LOBSTER`
- `SLatissima`
- `InstantRemineralisation`

PR #157, opened 2023-11-07
Bump Oceananigans to 0.90

jagoosw:
Some (internal) fixes required for light attenuation.


navidcy:
We need to update the example in the `README.md` + docs landing page.

jagoosw:
This is now failing from a bug in Oceananigans, I'll make a PR there and maybe just change how the temperature is set in this example for now (since it doesn't have any movement it can just be a function field instead).

jagoosw:
Actually changing this example to have a function field makes it quite messy since T is a tracer required by the bgc model so we would have to predefined all of the tracers like:
```
clock = Clock(; time = 0.0)
T = FunctionField{Center, Center, Center}(temp, grid; clock)
P = CenterField(grid)
N = CenterField(grid)

model = NonhydrostaticModel(..., tracers = (; N, P, T))
```

Which takes away from our demonstration of the automatic setup. I am going to try and run the example with a temperature forcing that is the derivative of the prescribed temperature and see if the time resolution is sufficient for it to work correctly.

jagoosw:
Should we merge #156 then this with a minor version update, then #155?

navidcy:
We can just do a minor update when all three are merged?

jagoosw:
Oh yeah I meant patch but we can just merge them all and do a minor release when #155 is merged

PR #162, opened 2023-12-08
Added passing of time to sediment tendency calculation

jagoosw:
I've come across an instance where I'm trying to write a different sediment model and need the forcings to have a time dependence so need to pass the time to `_calculate_tendencies!`, and modifies sediment tendency kernel to make new method definition possible.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OceanBioME"
 uuid = "a49af516-9db8-4be4-be45-1dad61c5a376"
 authors = ["Jago Strong-Wright <js2430@damtp.cam.ac.uk> and contributors"]
-version = "0.9.1"
+version = "0.9.2"
 
 [deps]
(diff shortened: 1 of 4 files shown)


navidcy:
@jagoosw my plate is a bit full now... But could try to look at this in a week or so..

navidcy (approved):
looks ok, I don't know how deep I can review this.
is there a test that tests this? or a validation that now works and before it wasn't?

jagoosw:
> looks ok, I don't know how deep I can review this. is there a test that tests this? or a validation that now works and before it wasn't?

Thank you for reviewing!

The reason I made these changes is that I'm using a sediment model (not BGC but this was the easiest interface for it) that depended on the surface radiation which I have as a time-dependent function, and I also realised it was more generally consistent with other tendency calculations that the time be included.

I then moved the kernels to be functions because (I think) you can't externally define new methods for `@kernel`s. If a user wanted to implement their own sediment model I think this would be essential.

PR #170, opened 2024-03-22
Added `invalid_fill_value` to `ScaleNegativeTracers`

jagoosw:
This PR adds a parameter `invalid_fill_value` to `ScaleNegativeTracers` so that when the total cell content is less than 0 it doesn't have to be filled with NaNs. While even less ideal this seems to be the easiest way to keep complex models positive (e.g. the near global model), and is akin to how lots of earth system models do it.

### DIFF
--- a/src/Utils/negative_tracers.jl
+++ b/src/Utils/negative_tracers.jl
@@ -39,8 +39,8 @@ struct ScaleNegativeTracers{FA, SA, FV, W}
     nan_fill_value :: FV
               warn :: W
 
-    ScaleNegativeTracers(tracers::FA, scalefactors::SA, warn::W, nan_fill_value::FV) where {FA, SA, FV, W} =
-        warn ? error("Warning not currently implemented") : new{FA, SA, W}(tracers, scalefactors, nan_fill_value, warn)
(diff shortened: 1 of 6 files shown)


src/Utils/negative_tracers.jl
```diff
@@ -60,14 +62,17 @@ github](https://github.com/OceanBioME/OceanBioME.jl/discussions/48).
 
 Future plans include implement a positivity-preserving timestepping scheme as the ideal alternative.
 
-If `warn` is true then scaling will raise a warning.
+~~If `warn` is true then scaling will raise a warning.~~
+
+`invalid_fill_value` specifies the value to set the total cell content to if the total is less than 0
+(meaning that total tracer conservation can not be enforced).
```
navidcy, line 68: perhaps we elaborate a bit more here... eg mention the caveat you included in the first comment in the PR? also, if other models follow similar practice perhaps mention 1-2 here?

navidcy (approved):
lgtm

only thing I think it'd be nice to mention the caveat in the docstring

jagoosw:
Thanks for sorting @navidcy, I'm going to bump the patch number and they merge

PR #179, opened 2024-07-11
Fixs seawater density conversion

jagoosw:
Was just wrong before, now it should be correct

density

### DIFF
--- a/src/Boundaries/gasexchange.jl
+++ b/src/Boundaries/gasexchange.jl
@@ -14,8 +14,6 @@ struct pCO₂{P0, P1, P2, PB, PW, FT}
               lower_pH_bound :: FT
               upper_pH_bound :: FT
                  boron_ratio :: FT
-           thermal_expansion :: FT
-          haline_contraction :: FT
 
 
     pCO₂(solubility::P0,
@@ -25,17 +23,14 @@ struct pCO₂{P0, P1, P2, PB, PW, FT}
          water_dissociaiton::PW,
          lower_pH_bound::FT,
          upper_pH_bound::FT,
-         boron_ratio::FT,
-         thermal_expansion::FT,
-         haline_contraction::FT) where {P0, P1, P2, PB, PW, FT} =
+         boron_ratio::FT) where {P0, P1, P2, PB, PW, FT} =
          new{P0, P1, P2, PB, PW, FT}(solubility, 
                                      bicarbonate_dissociation, 
                                      carbonate_dissociation, 
                                      boric_acid_dissociation, 
                                      water_dissociaiton, 
                                      lower_pH_bound, upper_pH_bound, 
-                                     boron_ratio, 
-                                     thermal_expansion, haline_contraction)
+                                     boron_ratio)
 end
 
 adapt_structure(to, pCO₂_model::pCO₂) = pCO₂(adapt(to, pCO₂_model.solubility),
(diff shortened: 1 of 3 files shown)


src/Boundaries/gasexchange.jl
```diff
@@ -45,17 +40,48 @@ adapt_structure(to, pCO₂_model::pCO₂) = pCO₂(adapt(to, pCO₂_model.solubi
...
 end
 
+"""
+    seawater_density(T, S)
+
+Returns the density of seawater at `T` °C and `S` psu in kg/m³ as
+per Chapter 5, Section 4.2 of Dickson, Sabine and Christian (2007, 
+http://cdiac.ornl.gov/oceans/Handbook_2007.html).
+
+With many thanks to Oscar Brandson for his implementation in cbsyst
```
johnryantaylor, line 57: There is a typo in Oscar's surname on line 57

src/Boundaries/gasexchange.jl
```diff
@@ -45,17 +40,48 @@ adapt_structure(to, pCO₂_model::pCO₂) = pCO₂(adapt(to, pCO₂_model.solubi
...
+(https://github.com/oscarbranson/cbsyst).
+"""
+@inline function seawater_density(T, S)
```
johnryantaylor, line 60: Should the seawater density also be a function of pressure?  What does cbsyst do for pressure?

jagoosw:
Going to do this all in one PR #186 186

src/Boundaries/gasexchange.jl
jagoosw: Maybe I should call it `surface_seawater_density`? `cbsyst` assumes that the units are per mass already so doesn't actually use the density calculation anywhere, but it also has no pressure factor.

johnryantaylor: Would it work in the future to use multiple dispatch to have seawater_density(T, S, P) that calculates the density with pressure?  In that case we can leave it how it is now.  Otherwise surface_seawater_density is probabliy better

glwagner: why not use TEOS-10?

jagoosw: That's a good idea I'll change the new PR to do that

jagoosw: Looking at the oceananigans code, is there a problem with the TEOS10 that means it's better to use the polynomial approximation?

glwagner: I mean, why not use `SeawaterPolynomials`, which provides a polynomial approxmation to TEOS10?

There's no issue with TEOS10, it's just expensive so we use polynomial approximations.

Even if you don't like the polynomial approximations currently in SeawaterPolynomials, I feel it's best to contribute any new polynomial approximations to that package (so they can be used everywhere) rather than here. See the package:

https://github.com/CliMA/SeawaterPolynomials.jl

jagoosw: Sorry should have updated here, I tried using SeawaterPolynomials as well as the TEOS peoples c-wrapper but then realised that it also isn't possibly to use it on GPU, so went with SeawaterPolynomials being the default.

I'm making this and a bunch of other changes in #186.

glwagner: Right, that is the purpose of SeawaterPolynomials

PR #183, opened 2024-07-12
Adds nanophytoplankton model

ciadht:
Added nanophytoplankton model


ciadht:
We believe we have coded up all the relevant parts of PISCES from the paper and included them in the modules. There remain a few issues which we have noted and left placeholders in their values. What remains to change in the PISCES documentation for this to run?

PR #184, opened 2024-07-12
Adds `IronPhosphate` sediment model

louis-de-neve:


src/Boundaries/Sediments/IronPhosphate.jl
```diff
@@ -0,0 +1,283 @@
...
+        @inline K_NO₂ = 10e-6 # M, Half–saturation constant for NO2
+        @inline K_Fe = 0.028 # wt-%, Half–saturation constant for Fe
+        @inline K_SO₄ = 0.1e-6 # M, Half–saturation constant for SO4
+        @inline K_TPO₄ = 10e-6 # M, Half–saturation constant for TPO4
+        #(:O₂, :NH₄, :NO₃, :NO₂, :N₂, :TPO₄, :FeOHP, :Feᴵᴵ, :FeS₂, :SO₄, :TH₂S, :CH₄, :TCO₂, :Gi)
+         
+        @inline kGi = 0.016 # day⁻¹, Rate constant for G0 degradation, Dale et al
+        @inline fT = 1 # TODO temperature correction for rates
+        @inline fₒₓ = 10 # Enhancement factor for POM degradation by O2
+        @inline per_day_to_per_seconds = 1 / (60 * 60 * 24)
```
jagoosw, line 212: btw if you do `using Oceananigans.Units` then you can just use units like below instead of using this variable you can write `max(whatever)/day`

PR #185, opened 2024-07-12
Make `BoxModel`s speedy again (+ introduce prescribed PAR)

jagoosw:
Various changes to make time stepping better (~10x speedup for time stepping part) and a new "output writer" (but don't have time to make it an actual `AbstractOutputWriter` at the moment) which is >10x faster for the really long file saving stage, resulting in ~30x speedup overall for outputting every 20 iterations.


jagoosw:
It has been pointed out to me that box models don't return auxiliary fields from `biogeochemical_auxiliary_fields` so I will add that in this PR

Issue #168, opened 2024-02-28
Update gas transfer coefficients


vtamsitt:
I'm interested in implementing updated gas transfer velocity (k) parameterisations (at the very least, Wanninkhof 1992 is outdated and could be replaced by [Wanninkhof 2014](https://aslopubs.onlinelibrary.wiley.com/doi/abs/10.4319/lom.2014.12.351)). It would be great to add other gas transfer velocities e.g. Ho et al. 2006, as is mentioned in the code comments. I'm guessing the best way to implement this would be to add all of these parameterisations (and more?) and include a kwarg when calling GasExchange? 
https://github.com/OceanBioME/OceanBioME.jl/blob/0b45a16ee2017af6ac3d8cebbbb613760983d6ae/src/Boundaries/gasexchange.jl#L127

johnryantaylor:
Hi @vtamsitt, thanks! That would be great and what you suggest seems like a good approach

vtamsitt:
Great, I'll work on this when I get a chance.

glwagner:
It could make sense to add a property called `transfer_velocity` here:

https://github.com/OceanBioME/OceanBioME.jl/blob/0b45a16ee2017af6ac3d8cebbbb613760983d6ae/src/Boundaries/gasexchange.jl#L142

then it'll be easy to add new models (for developers) or to switch between models (for users)

jagoosw:
Closed by #186

PR #186, opened 2024-07-15
Refactors gas exchange carbon chemistry (+ adds some complexity)

jagoosw:
This PR refactors the carbon chemistry calculations improving the clarity and supersedes #179.

It also moves both the carbon chemistry and gas exchange functions to `Models` since they are models, and separates them. 

This PR also implements the full carbon chemistry standards of the best practice (https://www.ncei.noaa.gov/access/ocean-carbon-acidification-data-system/oceans/Handbook_2007/Guide_all_in_one.pdf) so it can depend on Si and PO4, and approximated fluoride and sulfate. It also introduces proper (TEOS10 and polynomial approximations) density calculations.

Also, all of the equilibrium and pressure correction constants are also tested now.

I also introduce a function `calcite_saturation` that can compute `calcite_concentration` and saturation for use in other models.

This should also be faster since it removed a branching operation.

Finally, I changed the structure of `GasExchange` to make it more flexible (and correct, there was a bug in the computation of x(CO2) from fCO2).

Also, closes https://github.com/OceanBioME/OceanBioME.jl/issues/197, #200


jagoosw:
@bethcandish you could do some testing with the "best practice" parameters with this branch and see if it makes any difference, and if so change them here.

docs/src/model_components/air-sea-gas.md
```diff
@@ -151,12 +151,12 @@ resulting in the equilibrium constant:
 K_B = \frac{\ce{[B(OH)^-_4][H^+]}}{\ce{[B(OH)_3]}}.
 ```
 
-Finally, taking the DIC and Alkalinity in micro equivalents (i.e. scaled by ``10^{-3}/\rho_o`` from mmol C/m``^3``) denoted by ``\bar{DIC}`` and ``\bar{Alk}``, the carbonate alkalinity is given by ``Alk_C = \bar{Alk} + [B(OH)^-_4] + [OH^-]``, and we define the boron content, ``B``, to be ``R_{B:S}S`` mol/kg.
+Finally, taking the DIC and Alkalinity in mol / kg and equivilants / kg (i.e. scaled by ``10^{-3}/\rho_o``) denoted by ``\bar{DIC}`` and ``\bar{Alk}``, the carbonate alkalinity is given by ``Alk_C = \bar{Alk} + [B(OH)^-_4] + [OH^-]``, and we define the boron content, ``B``, to be ``R_{B:S}S`` mol/kg.
```
johnryantaylor, line 154: spelling should be equivalents, I think

src/Models/CarbonChemistry/carbonate_chemistry.jl
```diff
@@ -0,0 +1,107 @@
...
+                                      phosphate = 0,
+                                      upper_pH_bound = 14,
+                                      lower_pH_bound = 0)
+
+    ρₒ = seawater_density(T, S)
+
+    # Centigrade to kelvin
+    T += 273.15
+
+    # mili-equivilants / m³ to equivilants / kg
```
johnryantaylor, line 62: spelling should be "equivalents" throughout

johnryantaylor (commented):
The edits look good to me.  There are some comments in the carbon chemistry function that can be cleaned up.  It was a great idea to put the carbon ehcmistry as a separate model type - this will be very useful!

jagoosw:
@johnryantaylor do you think it would make sense to move carbon chemistry to be a property of `Biogeochemistry`?

Previously we had only used it in the CO2 flux but PISCES needs the calcite saturation for various things, and it makes sense to me that we separate it so it can be used for any model.

It wouldn't cause many changes and will just be there to be called, e.g. by the boundary condition.

jagoosw:
Also, I will update the docs page at some point but the code now includes the pressure correction for the rate constants. I have added some tests which ensure we have the constants right (when using the "best practice" parameters) like how Kgen does, so I think the only difference between the functionality here and cbsyst on the carbon front is the Mg and Ca bits.

jagoosw:
Final comment for now, I'm pretty sure that this PR isn't breaking since `OCMPI_default` wasn't part of the public API, but it's not clear to me if it counts as breaking because lots of things have moved around. I don't really want to make a breaking release because it was a lot of effort to bump the compact of all my packages that rely on it like the giant kelp stuff.

jagoosw:
I downloaded all of the latest GLODAP data (~200000 complete records) and get pH error to be 0.00018 ± 0.01173 with the "best practice" parameters (vs 0.006 ± 0.012 with the parameters from the NEMO source code).

image

Also, in the latest GLODAP, there are 9 cruises which report 10059 directly measured pCO2 values. With the "best practice" parameters I get an error of -7.6 ± 29.5 (vs 11 ± 28 for the NEMO values).

image

It also looks to me like the error variance is larger at lower temperatures and greater error at lower pH.

Given these results I think we should change the defaults to the "best practice" parameters.

jagoosw:
Oh dear GibbsSeaWater doesn't work on GPU

jagoosw:
Docs preview is done now and includes much better details of whats going on: https://oceanbiome.github.io/OceanBioME.jl/previews/PR186/model_components/carbon-chemistry/

docs/src/model_components/carbon-chemistry.md
```diff
@@ -0,0 +1,350 @@
...
+## Using the model
+To use the carbon chemistry model it is constructed by simply writing:
+```@example carbon-chem
+using OceanBioME
+
+carbon_chemistry = CarbonChemistry()
+```
+
+### Computing ``fCO_2`` and ``pH``
+To compute the fugacity of carbon dioxide (``fCO_2``) you call the model with the `DIC`, `Alk`alinity, `T`emperature, and `S`alinity:
```
johnryantaylor, line 18: Do we want to label DIC, Alk, T and S in input instead of relying on the argument order?  DIC and Alk have similar values, so using the order could cause confusion. I think that this would also allow us to use multiple dispatch to handle the case where pH is specified and we wouldn't have to send in NaN for DIC or Alk.

docs/src/model_components/carbon-chemistry.md
```diff
@@ -0,0 +1,350 @@
...
+```@example carbon-chem
+carbon_chemistry(2145, 2448, 25.4, 36.45)
+```
+This is sufficient when computing ``fCO_2`` at the surface, but if we wanted to know ``fCO_2`` at depth where there is higher pressure we can specify the pressure in bar like:
+```@example carbon-chem
+carbon_chemistry(2145, 2448, 25.4, 36.45; P = 100)
+```
+
+We may also be interested in the pH so we can request that be outputted:
```
johnryantaylor, line 27: Is it possible to output pCO2 and pH, or would the function need to be called twice?

johnryantaylor (commented):
These changed are great, thanks for your work on this!  I left a comment about the calling sequence.

docs/src/model_components/carbon-chemistry.md
jagoosw: This is a good idea and I'll change it now

docs/src/model_components/carbon-chemistry.md
jagoosw: We could do both but I'm not sure what the cleanest way to-do it is. 

I don't want to put any conditions like if X return Y elseif Z return A because the function has to run in a kernel and that can be slower on GPU (although if a kernel is doing this the branch will be the same for all of the threads so this might not actually cause a branching slowdown).

It could be:
```
    return ifelse(return_pCO₂ & return_pH, 
                  (pCO₂, pH),
                  ifele(return_pH, 
                        pH, 
                        ifelse(return_pCO₂,
                               pCO₂,
                               NaN)))
```

but that's kind of messy?

I do see the value of returning them both for diagnostics, but I don't think we'd ever want to use it for the prognostics because if I do e.g.
```
pH = carbon_chemistry(; DIC, Alk, T, S, return_pH=true)
pco2 = carbon_chemistry(; DIC, T, S, pH)
```
the value of `pco2` is different that if I do `carbon_chemistry(; DIC, Alk, T, S)` from the accuracy lost in logging and 10^ing the pH.

So to summarise, I think it would be useful diagnostically, but I can't think of a useful instance for it inside the models where the advantage of not having to call it twice would be the most important.

vtamsitt:
Great thanks, I'll go back and take another look at this. I'm not sure if it would be useful to have a way for users to input which wind product is used and then have the correct coefficient chosen from there, or whether to just include all the wind product-specific parameterizations and leave it up to the user.

> @vtamsitt the refactoring of the gas exchange model here should make adding different gas transfer parameterizations (#168) a lot easier because you can put anything in the `transfer_velocity` that you like as long as it has a method with arguments `u, T`.
> 
> I also changed the default to Ho et al., 2006 (which is the same as Wanninkhof, 2014 - https://x.com/_david_ho_/status/1414754481074589697) but it is important that the coefficients are wind product specific so you should use different values.

jagoosw:
Do you know if there is a list somewhere of the coefficients for each wind product? I can make a function that builds a `ScaledTransferVelocity` with the appropriate coefficient if the user specifies a wind product.

vtamsitt:
> @vtamsitt do you know if there is a list somewhere of the coefficients for each wind product? I can make a function that builds a `ScaledTransferVelocity` with the appropriate coefficient if the user specifies a wind product.

The [PySeaFlux](https://github.com/lukegre/pySeaFlux) [data product](https://zenodo.org/records/8280457) includes the coefficients they have calculated for  "ERA5", "JRA55", "NCEP1", "CCMP2" in SeaFlux.v2023.02_alpha_1982-2022.nc. These will presumably be recalculated each time the product is updated.

But these are for Wanninkhof 1992! 
The [code](https://github.com/lukegre/pySeaFlux/blob/master/pyseaflux/gas_transfer_velocity.py) includes a bunch of other gas transfer velocity parameterizations including Ho et al. 2006 and others but as far as I can tell haven't included coefficients for different wind products for the other parameterizations.

jagoosw:
If I understand correctly, the data product is a climatology of $k_w$ based on the Wanninkhof 1992 parameterization right? But I downloaded it and it doesn't include the parameter value they used so I assume they used the original values maybe.

In their code it says which wind products each parameterization is suitable for (e.g.  Ho 2006 scale factor says "The gas transfer velocity is for the QuickSCAT satellite wind product"). 

Since all of the parameterizations are some form of:
$$$
k_w = k_{660}(u) (Sc/660)^{-1/2}
$$$
where $f(u)$ is some polynomial (possibly piecewise e.g. `k_Li86`) parameterization I think I'll rename `ScaledTransferVelocity` to `SchmidtScaledTransferVelocity` which has `Sc` and `f` properties, and then I'll add in the default configurations.

vtamsitt:
> If I understand correctly, the data product is a climatology of kw based on the Wanninkhof 1992 parameterization right? But I downloaded it and it doesn't include the parameter value they used so I assume they used the original values maybe.
> 
> In their code it says which wind products each parameterization is suitable for (e.g. Ho 2006 scale factor says "The gas transfer velocity is for the QuickSCAT satellite wind product").


There's a separate file [SeaFlux.v2023.02_alpha_1982-2022.nc](https://zenodo.org/records/8280457/files/SeaFlux.v2023.02_alpha_1982-2022.nc?download=1) that contains the coefficient of gas transfer for the four wind products:

```
netcdf SeaFlux.v2023.02_alpha_1982-2022 {
dimensions:
        wind = 4 ;
variables:
         float alpha(wind) ;
                 alpha:_FillValue = NaNf ;
                 alpha:units = "(cm/hr) (m/s)^-2" ;
                 alpha:long_name = "coefficient of gas transfer" ;
                 string alpha:description = "The coefficient of gas transfer is scaled so that the global average kw is 16.5 cm/hr as recommended by Naegler (2009) for the period 1990-2019 using bomb 14-C estimates. Note that there is 20% uncertainty around the global average estimate of kw (16.5 ± 3.2 cm/hr).Note that during scaling, we calculate the area weighted average for kw over only the ice free ocean (ice frac < 5%) where pixels contain only ocean (sea frac > 95%). We also exclude the Black and Caspian sea from the globally averaged kw. " ;
         string wind(wind) ;
 data:
 
  alpha = 0.270875, 0.2601975, 0.2866424, 0.256789 ;
 
  wind = "ERA5", "JRA55", "NCEP1", "CCMP2" ;
 }
 (END) 
```

> 
> Since all of the parameterizations are some form of:$k_w = k_{660}(u) (Sc/660)^{-1/2}$ where f(u) is some polynomial (possibly piecewise e.g. `k_Li86`) parameterization I think I'll rename `ScaledTransferVelocity` to `SchmidtScaledTransferVelocity` which has `Sc` and `f` properties, and then I'll add in the default configurations.

This sounds good, thanks!

jagoosw:
Ah I see thank you, I've added the different parameterizations now and interestingly they're all very similar except Wanninkhof99 and McGillis01, and you can that Wanninkhof09 is quite differently parameterised from the others but is near by.

gas_transfer_velocities

jagoosw:
> alpha = 0.270875, 0.2601975, 0.2866424, 0.256789 ;
>  
>   wind = "ERA5", "JRA55", "NCEP1", "CCMP2" ;

It is interesting that these are all within the error estimates of Ho 2006 (and I can't find errors for other ones).

I've added them now too.

vtamsitt:
> Ah I see thank you, I've added the different parameterizations now and interestingly they're all very similar except Wanninkhof99 and McGillis01, and you can that Wanninkhof09 is quite differently parameterised from the others but is near by.
> 
> gas_transfer_velocities

Thanks for adding all these. This is really interesting, thanks!

jagoosw:
TODO: check model_implemntation page when rendered because it shouldn't have been using the forcing fallback

validation/glodap_fit.jl
```diff
@@ -0,0 +1,143 @@
...
+
+    density = density_function(T, S, P, lon, lat)
+
+    DIC = data[n, DIC_name] * density * 1e-3
+    Alk = data[n, Alk_name] * density * 1e-3
+    
+    silicate = data[n, Si_name] * density * 1e-3
+    phosphate = data[n, PO_name] * density * 1e-3
+
+    silicate = ifelse(silicate == -9999, 0, silicate)
```
chloeyhuang, line 84: the ifself statement should probably go before the units of silicate and phosphate are converted?

PR #189, opened 2024-07-25
Makes testing more robust

jagoosw:
There is a race condition in the testing pipeline where if you start a job while another is running, the first might delete the julia instance before the second one completes.

This PR should prevent that condition from occuring.


jagoosw:
This seems to have worked here, I re-ran this + another of the open PRs tests at the same time and it didn't cause any issues like it normally would

jagoosw:
And it segfaulted

jagoosw:
As concluded in the Oceananigans PR I made this doesn't seem to work

PR #190, opened 2024-07-26
Attempts to make tests work better using the buildkite plugin

jagoosw:
Not making this a draft or the CI won't run

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -35,7 +35,7 @@ steps:
     depends_on: "init_gpu"
     notify:
       - github_commit_status:
-          context: ":speedboat: GPU tests"
+          context: "GPU tests"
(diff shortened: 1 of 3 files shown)


jagoosw:
So this may be working now, so I'm going to make a branch from this branch and PR it into main, then make them both run at the same time to see if that works as a stress test

jagoosw:
This is way way slower for some reason? In #186 (old version of pipeline) the tests finish in <20min but this is still running after 45min.

jagoosw:
Maybe I should experiment with what we're caching

jagoosw:
Ah it realises that these branches are duplicates I think

jagoosw:
This does seem a lot more robust now but also a lot slower, I will investigate adding more caching

jagoosw:
I don't think we should do it this way since its so slow

PR #205, opened 2024-08-17
Adds `MultiBandPhotosyntheticallyActiveRadiation`

jagoosw:
This PR adds `MultiBandPhotosyntheticallyActiveRadiation` with default parameters replicating the "Morel" model used in other implementations of PISCES.

It required a small PR into Oceananigans to make it work (https://github.com/CliMA/Oceananigans.jl/pull/3683) which is now merged and just waiting for a release, so this PR currently changes to use Oceananigans main branch instead for now (but this should **not** be merged into main).


jagoosw:
I'm going to merge this with main when the oceananigans patch is out, and then you can pull main into the PISCES branch because that will probably be less confusing

PR #212, opened 2024-09-13
Refactors to improve code clarity and make changing parameterisations easier

jagoosw:
I think there is still some room to make this more flexible, for example changing `nanophytoplankton` and `diatoms` to just `phytoplankton` which is a `NamedTuple` of phytoplankton classes, but then we need functions like `ammonia_uptake(all_the_phytoplanktons)`. This would be useful for experimentation in the future with different things but I am not going to implement here as we just want a working version, and then when people want to experiment that can make these kind of changes.

To do:

- [x] docstrings
- [x] show methods
- [ ] check for any more deviations from Aumount 2015
- [x] Fix negative tracer scaling
- [x] comment all parameters units
- [x] documentation


jagoosw:
This all looks good now except that oxygen is coming from no where 
image

jagoosw:
Changes from Aumount 2015 to match NEMO:
- enhanced quadratic mortality rate for diatoms changes from ``w_0 + w_1 (1-L)`` to ``w_0 + 0.25 w_1 (1 - L^2)/(0.25 + L^2)``
- P-I slope can vary from ``alpha``, but default behaviour is to not enhance with depth, also diatom parameterization always returns the same as nano so ignoring it

jagoosw:
Everything seems reasonable now, except I think there must be a small leak of something because there is a very slow trend up/down in the primary production-related tracers

jagoosw:
Chlorophyll uptake is documented incorrectly

jagoosw:
So, I think chlorophyll production should be given as:
```math
\frac{\partial IChl}{\partial t} += 12 (\theta^{min} + \rho(\theta^{max}-\theta^{min}))\mu I,
```
```math
\rho = \left[L_{day}\frac{\mu I}{f_1}\right]L\frac{12P}{PChl}\frac{L_P}{\alpha PAR},
```
which makes ``\rho`` unitless.

jagoosw:
``\theta_{opt}`` for iron in the code is in `mol Fe / mol C` so converting to `\mu mol Fe / mmol C` by multiplying by a factor of ``10^3``

jagoosw:
image

jagoosw:
Once the tests pass I'd like to merge this back into the PISCES branch, is that okay with you @ciadht and @hannahmw1

jagoosw:
So I had todo some horrible things to get this to work on GPU because it seems that `_apply_iterate` is not implemented in CUDA C or CUDA.jl or somewhere along the line for very long tuples, so this code also introduces a hack which I will hopefully implement in Oceananigans.

jagoosw:
I think there is a whole new level of abstraction we could do by changing the model to be discrete form, which will likely make this run more efficiently, as well as facilitate grouping the phytoplankton etc. as discussed above.

Issue #214, opened 2024-09-20
Data assimilation example is broken!!!


jagoosw:
https://oceanbiome.github.io/OceanBioME.jl/stable/generated/data_assimilation/

jagoosw:
Closed by #215

PR #215, opened 2024-09-20
Fixs data assimilation example

jagoosw:
Solves #214


jagoosw:
Definitly not fixed

jagoosw:
This is correct now

PR #217, opened 2024-09-24
Rewrites PISCES to be a discrete form model, I think over all this makes the implementation a lot more flexible

jagoosw:
Even if it was a huge pain to basically rewrite for a third time...


jagoosw:
think I've broken the particles

jagoosw:
doesn't fail locally, not sure why its failing now

PR #223, opened 2024-10-22
adds field name to particle interpolation

jagoosw:


src/Particles/tracer_interpolation.jl
glwagner, line 1: This whole file seems to implement general Oceananigans utilities. It might be good to have a think about what belongs in OceanBioME vs Oceananigans.

The reason to put things in Oceananigans is because it will become available to a broader community of users. In turn, those users will test, provide feedback, and further develop the code, adding functionality and ensuring its quality. This provides mutual benefits to OceanBioME and its users.

jagoosw: This should probably be in Oceananigans too:
https://github.com/OceanBioME/OceanBioME.jl/blob/26483ba13e6385d7dd3f47a54df586c40acb566e/src/Particles/atomic_operations.jl

PR #225, opened 2024-10-23
Update to use Oceananigans v0.92.1

ali-ramadhan:
As of Oceananigans v0.92.1 (and PR https://github.com/CliMA/Oceananigans.jl/pull/3732), `TracerAdvection` is now called `FluxFormAdvection` so this PR just updates the code.

I thought it warranted tagging v0.93.0 but I guess `TracerAdvection` was not exported so it's not technically a breaking change in terms of the public interface.


jagoosw (approved):
Thank you for sorting this out! We should change the version and release a patch here too

ali-ramadhan:
Good idea! I just bumped v0.13.1. Will merge once tests (hopefully) pass.

PR #244, opened 2025-01-24
abstract out biogeochemical particles

jagoosw:
This allows a particle model to bypass all of the stuff which makes simple particle models easy to setup, and I need it for this model: https://github.com/jagoosw/GiantKelpDynamics/pull/7

Maybe BiogeochemicalParticles should be renamed ContinuousBiogeochemicalParticles

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OceanBioME"
 uuid = "a49af516-9db8-4be4-be45-1dad61c5a376"
 authors = ["Jago Strong-Wright <js2430@damtp.cam.ac.uk> and contributors"]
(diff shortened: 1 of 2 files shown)


johnryantaylor (approved):
LGTM.  The name ContinuousBiogeochemicalParticles is confusing to me, though, since particles are inherently discrete (in space).

jagoosw:
> LGTM. The name ContinuousBiogeochemicalParticles is confusing to me, though, since particles are inherently discrete (in space).

That's a good point it doesn't really make sense in the same way as `Discrete` and `Continous` bgc, I think I'll leave it for now.

Issue #248, opened 2025-03-11
README.md needs to add CairoMakie to make it self contained


dorchard:
The example in the README.md seems to need the CairoMakie package in order to create the plot, but this dependency isn't explicit so the `record` function is not defined, e.g., producing error:
```
ERROR: UndefVarError: `record` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ ~/.../example.jl:72
```

PR #249, opened 2025-03-11
Add CairoMakie dependency into README example

dorchard:
This resolves the missing dependency to produce the visualisation in the readme. See https://github.com/OceanBioME/OceanBioME.jl/issues/248

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,11 @@ julia> Pkg.add("OceanBioME")
 As a simple example lets run a Nutrient-Phytoplankton-Zooplankton-Detritus (NPZD) model in a two-dimensional simulation of a buoyancy front. This example requires Oceananigans, so we install that first:
 
 ```julia
-using Pkg; Pkg.add("Oceananigans")
+using Pkg; Pkg.add("Oceananigans"); Pkg.add("CairoMakie")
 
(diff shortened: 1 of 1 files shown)


dorchard:
This makes the example then work for me!

jagoosw:
Thank you for sorting this! It might be better if we moved it into the hidden visualization bit here so the model and visualization bits are all separate?
https://github.com/OceanBioME/OceanBioME.jl/blob/4237d06a55d1bbd670346160d25b8ac2780630e9/README.md?plain=1#L65-L70

I'll sort out the documentation failing, I need to merge https://github.com/OceanBioME/OceanBioME.jl/pull/247 first.

dorchard:
Yes good idea - I've moved those two lines to be with the visualisation code.

dorchard:
Hm I don't understand why the CI failed here. Is the merge of this PR pending fixing this issue with the CI?

jagoosw:
Sorry I compleltly forgot to merge this

jagoosw:
This CI problem is a known issue, I'll resolve conflicts and it should be fixed upstream

jagoosw:
Oh it looks like this is already fixed upstream, thank you for contibuting!

PR #250, opened 2025-03-20
Makes carbon chemistry type stable

jagoosw:
CC: @johnryantaylor


Project.toml
johnryantaylor, line 1: I don't think we need Measurements in the dependency list, do we?

johnryantaylor (changes requested):

Project.toml
jagoosw: oh no this was a mistake

johnryantaylor: Ok, great, I think that with that removed it will pass the tests and be ready to merge.

johnryantaylor: I removed the Measurements lines from Project.toml

johnryantaylor (approved):

Project.toml
johnryantaylor: Hi @jagoosw, there are some merge conflicts that need to be sorted out on this one.  Can you have a look?

johnryantaylor:
Thanks @jagoosw!  I have also noticed that it isn't type stable when returning pH.  I have been planning to look into that but haven't had a chance to do that yet.  It might be coming from the solve for H+, but I'm not sure.

PR #255, opened 2025-04-25
Makes carbon chemistry type stable made type stable attempt 2

jagoosw:
Supercedes #250 

CC: @johnryantaylor

### DIFF
--- a/src/Models/CarbonChemistry/carbon_chemistry.jl
+++ b/src/Models/CarbonChemistry/carbon_chemistry.jl
@@ -206,7 +206,7 @@ end
 solve_for_H(pH::FT, args...) where FT = convert(FT, 10.0) ^ - pH
 
 solve_for_H(::Nothing, params, upper_pH_bound::FT, lower_pH_bound) where FT =
-    find_zero(alkalinity_residual, (convert(FT, 10.0) ^ - upper_pH_bound, convert(FT, 10.0) ^ - lower_pH_bound), Bisection(); atol = convert(FT, 1e-10), p = params)
+    find_zero(alkalinity_residual, (convert(FT, 10.0) ^ - upper_pH_bound, convert(FT, 10.0) ^ - lower_pH_bound); atol = convert(FT, 1e-10), p = params)
 
 # display
 summary(::IO, ::CarbonChemistry) = string("`CarbonChemistry` model")
--- a/src/Models/CarbonChemistry/carbon_chemistry.jl
+++ b/src/Models/CarbonChemistry/carbon_chemistry.jl
@@ -142,31 +142,31 @@ unless `pH` is specified, in which case intermediate computation of `pH` is skip
 
 Alternativly, `pH` is returned if `return_pH` is `true`.
 """
-@inline function (p::CarbonChemistry)(; DIC, T, S, Alk = 0, pH = nothing,
+@inline function (p::CarbonChemistry)(; DIC::FT, T, S, Alk = zero(DIC), pH = nothing,
                                         P = nothing,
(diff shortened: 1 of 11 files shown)


johnryantaylor:
Hi @jagoosw, thanks for the updates on this.  When I run the code below, I am still getting pH as a Float instead of a measurement.  It is possible that I don't have the PR correctly loaded.  Can you check if this works for you?

using OceanBioME
using Measurements
carbon_chemistry = CarbonChemistry()
DIC = 2100.0 ± 1.0
Alk = 2300.0 ± 1.0
T = 12 ± 0.01
S = 35 ± 0.01
pH = carbon_chemistry(DIC = DIC, Alk = Alk, T=T, S=S, return_pH=true)

jagoosw:
> Hi @jagoosw, thanks for the updates on this. When I run the code below, I am still getting pH as a Float instead of a measurement. It is possible that I don't have the PR correctly loaded. Can you check if this works for you?
> 
When I run this I get a measurement:
```julia

julia> using OceanBioME

julia> using Measurements

julia> carbon_chemistry = CarbonChemistry()
`CarbonChemistry` model which solves for pCO₂ and pH

julia> DIC = 2100.0 ± 1.0
2100.0 ± 1.0

julia> Alk = 2300.0 ± 1.0
2300.0 ± 1.0

julia> T = 12 ± 0.01
12.0 ± 0.01

julia> S = 35 ± 0.01
35.0 ± 0.01

julia> pH = carbon_chemistry(DIC = DIC, Alk = Alk, T=T, S=S, return_pH=true)
8.049355360693447 ± 0.0
```

(although it's a bit suspicious that it's $\pm 0$?)

To load this branch you should just have todo:
```julia
pkg> add https://github.com/OceanBioME/OceanBioME.jl#jsw-jrt/make-carbon-chemistry-type-stable-take-2
```

johnryantaylor:
Hi @jagoosw, I have looked into this a bit more, and the problem is coming from the find_zero function.  If we simply remove Bisection(), then it works.  The Bisection method doesn't seem to work for non floating point variables.  The A42 method is the default for non floating point types, and it looks like it preserves the measuremnt type without zeroing the error.

src/Models/CarbonChemistry/carbon_chemistry.jl
```diff
@@ -203,10 +203,10 @@ Alternativly, `pH` is returned if `return_pH` is `true`.
 end
 
 # solves `alkalinity_residual` for pH
-solve_for_H(pH, args...) = 10.0 ^ - pH
+solve_for_H(pH::FT, args...) where FT = convert(FT, 10.0) ^ - pH
 
-solve_for_H(::Nothing, params, upper_pH_bound, lower_pH_bound) =
-    find_zero(alkalinity_residual, (10.0 ^ - upper_pH_bound, 10.0 ^ - lower_pH_bound), Bisection(); atol = 1e-10, p = params)
```
johnryantaylor, line 209: Remove Bisection() on this line

glwagner: also careful about hardcoded Float64!

johnryantaylor: actually this is the line from the old code which isn't type stable.  The new code should now be type stable.

jagoosw:
Okay, that sounds good and it returns a non-zero error for me now too

jagoosw:
Btw the tests are passing for this its just build kite has stopped reporting the status for some reason https://buildkite.com/oceanbiome/oceanbiome/builds/813

So this is ready to merge when someone approves it.

Issue #270, opened 2025-07-17
Update Oceananigans version


MarionBWeinzierl:
Update the Oceananigans version used in the tests and in OceanBioME in general to 0.97.

Also update the documentation.

PR #271, opened 2025-07-17
Examples after oceananigans v96

Mikolaj-A-Kowalski:
Addresses #270 

Updates the allowed version of Oceananigans to v0.96 and v0.97. 

In addition:
- updates the code examples to take account of the Output writers renaming in v0.96.
- add the same Oceananigans version specifiers in core package and in doc

I have not touched the code in `benchmarks` and `paper`. The former would require more modifications to run again, why I assumed the latter is better left as it is for reproducibility (?). Let me know If I should try to change either as well. 

I will also try to see if it would be possible to remove explicit dependencies in the `doc` on the packages that are implicit from `OceanBioMe`.


Mikolaj-A-Kowalski:
I have added an additional commit. I *think* that removing explicit version requirements in the `docs` subpackage should resolve the danger of using different versions of shared dependencies between `test` and `doc` jobs. 

Basically if the `docs` does not apply any constraints the only constraints on the `OceanBioME` dependencies should be the ones it applies (... the problem may still arise due to constraints from new dependencies e.g. `Documenter`, but it shouldn’t impact key packages i.e. `Oceananigans` )

johnryantaylor:
Thanks @Mikolaj-A-Kowalski, this sounds like a good idea to me.  I agree that the paper folder should stay linked with the version of Oceananigans that was used in the paper for reproducibility purposes.

Mikolaj-A-Kowalski:
> It failed the docs one.

I have seen it ;-) 
Will try to find out and fix what is going wrong this morning. 

Then we will see if the status will be reported for the passing job...

Mikolaj-A-Kowalski:
> It failed the docs one. 

I think I got it. Pushed a separate a commit with a fix. (Will need to get squashed before merge though).
It fixed the problem on my machine. We will see if the BuildKite agrees :crossed_fingers:  

Curiously the problem was not as trivial as I hoped. It was caused by a regression in Oceananigans from [here](https://github.com/CliMA/Oceananigans.jl/commit/8a8e4aa1417958e880a037549db2185a5de8de1d). Introduced in version `0.9.28`. 

Long story short the `update_boundary_condition!` method got split into two: 
- `update_boundary_condition!`
- `update_boundary_conditions!` 

It was not considered a breaking change I guess since these are technically Oceananigans internals. However OceanBioMe is so closely integrated that these *non-breaking* changes are breaking.

Mikolaj-A-Kowalski:
Docs were still failing due to `WENO` constructor containing grid :
```diff
-WENO(;grid)
+WENO()
```
Seems to fix the failure, but I have not been able to track down where does it originate from. As in what has changed in what version to cause it to become outdated.

AdelekeBankole (approved):
Thank you @Mikolaj-A-Kowalski, this looks good to me.

ali-ramadhan:
Thank you @Mikolaj-A-Kowalski! Would it be possible to tag v0.14.2 as part of this PR since it adds support for Oceananigans v0.97?

tomchor:
Hey, all, I just wanted to check the status here. I'd like to use OceanBioME for some development (for which I need Oceananigans 0.97) and I've been kinda having to hack Manifests together from different branches. It'd be helpful to have this merged, but it seems like tests are stuck somehow?

jagoosw:
Hi @tomchor,

I'll merge this if thats okay with @Mikolaj-A-Kowalski?

Mikolaj-A-Kowalski:
> Hi @tomchor,
> 
> 
> 
> I'll merge this if thats okay with @Mikolaj-A-Kowalski?
> 
> 

Yes it is fine with me.
Sorry for the late response, I am on holiday at the moment.

PR #272, opened 2025-07-21
Add support for Oceananigans v0.97

ali-ramadhan:


ali-ramadhan:
I suspect this PR will require #271 for examples to pass.

jagoosw:
I think #271 adds this support too right?

PR #281, opened 2025-08-28
Adds nothing method for `compute_flux_bc_tendencies!`

jagoosw:
This PR adds a method for `compute_flux_bc_tendencies!` for `BoxModel` to maintain box model functionality after Oceananigans 0.97.6.

I think we shouldn't need to enforce compatability >0.97.6 since I think lower versions will just give a warning that `compute_flux_bc_tendencies!` can't be imported.

@johnryantaylor


jagoosw:
Superseded by #278

PR #286, opened 2025-09-08
Attempts to fix CI concurrency problem

jagoosw:
This PR updates our buildkite process to use `juliaup` and different directories for each build (as Oceananigans does) to stop concurrent jobs interfering with each other (e.g. see https://buildkite.com/oceanbiome/oceanbiome/builds/879 and https://buildkite.com/oceanbiome/oceanbiome/builds/880). 

In the interest of speeding up documentation building I have also:

- moved the eady example to run on GPU (since the benchmarking would suggest its large enough to be sped up
- removed the `data_forced` example since it took ~20 minutes vs 5min for all of the others
- moved the example building to separate buildkite steps so that they occur in parallel, then `docs/make.jl` runs after 

It now takes around 15min for the docs to build vs 50min before

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OceanBioME"
 uuid = "a49af516-9db8-4be4-be45-1dad61c5a376"
 authors = ["Jago Strong-Wright <js2430@damtp.cam.ac.uk> and contributors"]
-version = "0.14.4"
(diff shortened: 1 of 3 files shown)


johnryantaylor:
Yes, no problem to remove it.  That example is a bit historical now (similar to your first paper).

PR #291, opened 2025-09-18
Make it run only on one agent

jagoosw:
Does some funny business with the pipeline uploading to try and get it to only run on one agent.

https://buildkite.com/docs/agent/v3/cli-start#run-a-job-on-the-agent-that-uploaded-it-also-known-as-node-affinity


jagoosw:
I have also changed it so that buildkite-agent runs with Nice=10 on kelp and ocean so user jobs have higher priority. This shouldn't normally effect the CI but if someone is running something on them it might slow down the CI a bit.

PR #305, opened 2025-10-23
0.15.0 Make `CarbonChemistry` be able to return pCO2

jagoosw:
This changes the `CarbonChemistry` API to have a kwarg `output` which by default is `Val(:fCO₂)` or can be `Val(:pCO₂)` or `Val(:pH)`. Also adds units to docstring.

@johnryantaylor @j-ostiguy


src/Models/CarbonChemistry/carbon_chemistry.jl
```diff
@@ -152,15 +164,15 @@ Alternativly, `pH` is returned if `return_pH` is `true`.
...
+    ρₒ = p.density_function(T, S, ifelse(isnothing(P), one(DIC), P), lon, lat)
 
     # Centigrade to kelvin
     T += convert(FT, 273.15)
 
     # mili-equivalents / m³ to equivalents / kg
     Alk *= convert(FT, 1e-3) / ρₒ
 
-    # mmol / m³ to mol / kg
+    # mmol / m³ to mol / kgg
```
johnryantaylor: what is "kgg"? Is this a typo?

docs/src/model_components/carbon-chemistry.md
```diff
@@ -28,9 +28,9 @@ This is sufficient when computing ``fCO_2`` at the surface, but if we wanted to
 carbon_chemistry(; DIC, Alk, T, S, P = 100)
 ```
 
-We may also be interested in the pH so we can request that be outputted:
+We may also be interested in the total pH so we can request that be outputted:
```
johnryantaylor: It would be useful to add that "When output=Val(:variable) is specified, the named variable is outputted instead of fCO2."  The user might be expecting fCO2 _and_ the named variable to be outputted.

PR #308, opened 2025-10-29
0.14.10 Improve root finding (sugar kelp and carbon chem) by computing analytical derivatives and using newton raphson solver

jagoosw:
Inhouse newton solve (since we can work out the derivative) instead of using `find_zero` bisection. 3x speed improvement to `photosynthesis(kelp, T, PAR)`

CC: @zhennaa

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,6 @@
 include("dependencies_for_runtests.jl")
 
+include("test_solvers.jl")
 include("test_utils.jl")
 include("test_light.jl")
 include("test_particles.jl")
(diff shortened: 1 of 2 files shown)


jagoosw:
I was having a problem with getting plain newton raphson to converge for the whole parameter space for the carbon chemistry (DIC/Alk/T/S) so have now implemented a few solvers (NewtonRaphson/NewtonRaphsonSafe (Newt Safe from numerical recipies), bisection, and damped newton raphson). I found that the newt safe was a lot slower but that the the damped newton was only fractionally slower than plain newton raphson across most of the parameter space (and converged for the whole space). So now the default is plain newton raphson for the kelp and damped for the carbon chemistry.

PR #312, opened 2025-11-25
0.16.2 Use Julia 1.12 for Buildkite tests + fixes

ali-ramadhan:
Seeing if all tests pass on Julia 1.12. Since Oceananigans v0.101 now supports and only runs tests on Julia 1.12 it might make sense to upgrade OceanBioME.jl to Julia 1.12.

PR #315, opened 2025-11-26
Reintroduce support for Julia 1.10/1.11

jagoosw:
@ali-ramadhan I don't think this was an intentional change in #312? Oceananigans still supports down to 1.9.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OceanBioME"
 uuid = "a49af516-9db8-4be4-be45-1dad61c5a376"
+version = "0.16.3"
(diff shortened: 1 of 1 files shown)


jagoosw:
> Ah sorry yes this was my mistake. Thanks for fixing!
> 
> Should it be `julia = "1.9"` like in https://github.com/CliMA/Oceananigans.jl/blob/6b1c59bad589311774d1e9aeb8fb1fbbb4e3f126/Project.toml#L107 ?

I'm not sure why it was 1.10 before, maybe we should support 1.9 also

PR #323, opened 2025-12-10
Add buffer variables for the Biogeochemical Particles

### LINKED ISSUE #321: Allow to share computation between kernels in particle models
By looking at the profiling results for the runs of LOBSTER model with large number of `SugarKelp` particles  (range of $2^{18} - 2^{20}$) the main extra contribution to the runtime comes from the [`photosynthesis`](https://github.com/OceanBioME/OceanBioME.jl/blob/4667bd10f8db10906b4674a79de9995e41992bdf/src/Models/Individuals/SugarKelp/equations.jl#L92) function which involves expensive Newton iteration. 

Furthermore, it appears that the heavy `photosynthesis` calculation is performed 6 times (in `_update_traced_tendencies!` kernel for `:DIC`, `:O_2`, `:DOC` and `:DON` tracers and in `_tendency_kernel!` for `:N` and `:C`). As far as I am able to tell the inputs come from current state of the fields and hence do not change between invocations of the  `_update_traced_tendencies!` and  `_tendency_kernel!` for different tracers. Hence, it appears that the same calculation is performed multiple times.  This is the impression I got from looking at the code, but need o verify that this is indeed a case by logging values from the run.    

If the above is correct, it would be beneficial to refactor 'Individual' models to give them ability to share some result across multiple kernels. I imagine we could allow them to declare some `buffer` space for each particle, which will be filled by new type of kernel run before  `_update_traced_tendencies!` and  `_tendency_kernel!`. Then when updating tendencies the `photosynthesis` result could just be read from memory 🤞

Mikolaj-A-Kowalski:
Resolves #321 

Basically it allows a biogeochemical model in the particles to declare buffer for some heavy-to-compute (i.e. `photosynthesis` function result) variables so they can be reused across multiple kernels that update tendencies for the tracers and particle fields. 

In contrast to what I said in #321, the buffers need to be filled twice since particle move between tracer and fields tendencies are updated. 

It is still a draft so we can discuss alternative/better way to introduce buffers. In particular the following points: 
- At the moment there is no differentiation between a buffer variables  for update of tracer and field tendencies, but in principle different set of variables may be needed. Should we try to have separate `buffer_variable` registration and filling for these steps? 
- I wasn't sure where the kernel that does the filling should be defined. At the moment it stayed in `Particle` and just delegates to a function provided by a specific `Model/Individual`. But I was wondering if the kernel should not be defined by a specific model so it can be optimised/tailored more to calculation of the specific buffer variable... 

The performance change will follow in the post shortly. 

TODO: 
- [x] add unit tests for buffers

### DIFF
--- a/src/Models/Individuals/SugarKelp/buffer.jl
+++ b/src/Models/Individuals/SugarKelp/buffer.jl
@@ -1,0 +1,4 @@
+import OceanBioME.Particles: compute_buffer_variable
+
+compute_buffer_variable(::Val{:Photosynthesis}, kelp::SugarKelp, u, v, w, T, NO₃, NH₄, PAR) =
+    photosynthesis(kelp, T, PAR)
--- a/src/Models/Individuals/SugarKelp/coupling.jl
+++ b/src/Models/Individuals/SugarKelp/coupling.jl
@@ -1,43 +1,41 @@
 #:NO₃, :NH₄, :DIC, :O₂, :DOC, :DON, :bPOC, :bPON
 
-@inline function (kelp::SugarKelp)(::Val{:NO₃}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR)
+@inline function (kelp::SugarKelp)(::Val{:NO₃}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR, photosynthesis_var)
     J = nitrate_uptake(kelp, N, NO₃, u, v, w)
 
     return -J * A / (day * 14 * 0.001) # gN/dm^2/hr to mmol N/s
 end
 
-@inline function (kelp::SugarKelp)(::Val{:NH₄}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR)
+@inline function (kelp::SugarKelp)(::Val{:NH₄}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR, photosynthesis_var)
     J = ammonia_uptake(kelp, t, A, N, C, T, NH₄, u, v, w)
     
     return -J * A / (day * 14 * 0.001) # gN/dm^2/hr to mmol N/s
 end
 
-@inline function (kelp::SugarKelp)(::Val{:DIC}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR)
-    P = photosynthesis(kelp, T, PAR)
+@inline function (kelp::SugarKelp)(::Val{:DIC}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR, photosynthesis_var)
 
     μ = growth(kelp, t, A, N, C, T, NH₄, u, v, w)
 
     R = respiration(kelp, t, A, N, C, T, NO₃, NH₄, u, v, w, μ)
     
-    return -(P - R) * A / (day * 12 * 0.001)  # gC/dm^2/hr to mmol C/s
+    return -(photosynthesis_var - R) * A / (day * 12 * 0.001)  # gC/dm^2/hr to mmol C/s
 end 
 
 # I now know that this may not be correct..., it probably should vary by where the nitrogen comes from
-@inline (kelp::SugarKelp)(::Val{:O₂}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR) =
-    - kelp(Val(:DIC), t, A, N, C, u, v, w, T, NO₃, NH₄, PAR)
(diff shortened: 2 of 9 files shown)


Mikolaj-A-Kowalski:
In terms of performance of the total time executing the 'update' kernels (both tracer and field tendencies) for $2^{18}$ particles decreases from ~ 33 ms to 17ms (~50%)

Looking in Nsight sytems this is the old:

image

This is the new: 
image

Basically the majority of the time is now concentrated in the `fill_single_buffer_variable` kernel that does the photosynthesis calculation. Further if we look a bit up in the report to the 'SM Wrap Occupancy' section we will see that we have a significant problem related to load 'last-wave' problem. The kernel starts at its theoretical occupancy of 50%, but it drops almost immediately ending at sad 3% for the last third of the kernel execution :cry:  

I guess this is just related to the load imbalance in the Newton iteration. The good news is that it probably can be significantly optimised delivering even more gain. 

The only cost of the change is extra memory (single double) per particle that needs to be stored on the GPU. I imagine though that the memory cost of the particles is much smaller than tracers in practice so it is not too heavy price to pay. 

I attach the original NSight reports:  [reports.tar.gz](https://github.com/user-attachments/files/24084908/reports.tar.gz)

The benchmark script: [LOBSTER_3D_profile_particles.jl.txt](https://github.com/user-attachments/files/24084936/LOBSTER_3D_profile_particles.jl.txt)

src/Models/Individuals/SugarKelp/buffer.jl
```diff
@@ -0,0 +1,5 @@
+import OceanBioME.Particles: compute_buffer_variable
+
+function compute_buffer_variable(::Val{:Photosynthesis}, kelp::SugarKelp, u, v, w, T, NO₃, NH₄, PAR)
+    return photosynthesis(kelp, T, PAR)
+end
```
jagoosw:
```suggestion
compute_buffer_variable(::Val{:Photosynthesis}, kelp::SugarKelp, u, v, w, T, NO₃, NH₄, PAR) =
    photosynthesis(kelp, T, PAR)
```

src/Models/Individuals/SugarKelp/coupling.jl
```diff
@@ -1,19 +1,18 @@
 #:NO₃, :NH₄, :DIC, :O₂, :DOC, :DON, :bPOC, :bPON
 
-@inline function (kelp::SugarKelp)(::Val{:NO₃}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR)
+@inline function (kelp::SugarKelp)(::Val{:NO₃}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR, P)
```
jagoosw:
```suggestion
@inline function (kelp::SugarKelp)(::Val{:NO₃}, t, A, N, C, u, v, w, T, NO₃, NH₄, PAR, photosynthesis)
```

I in this context its a bit confusing to use `P` since `P` almost always refers to phytoplankton

src/Particles/Particles.jl
```diff
@@ -22,23 +22,25 @@ import Adapt: adapt_structure
...
 
-struct BiogeochemicalParticles{N, B, A, F, T, I, S, X, Y, Z} <: AbstractBiogeochemicalParticles
+struct BiogeochemicalParticles{N, B, A, F, T, I, S, BufferType, X, Y, Z} <: AbstractBiogeochemicalParticles
       biogeochemistry :: B
             advection :: A
                fields :: F
           timestepper :: T
   field_interpolation :: I
          scalefactors :: S
+               buffer :: BufferType
```
jagoosw:
```suggestion
struct BiogeochemicalParticles{N, B, A, F, T, I, S, BT, X, Y, Z} <: AbstractBiogeochemicalParticles
      biogeochemistry :: B
            advection :: A
               fields :: F
          timestepper :: T
  field_interpolation :: I
         scalefactors :: S
               buffer :: BT
```

src/Particles/fill_buffer.jl
```diff
@@ -0,0 +1,29 @@
...
+
+    dev = device(architecture(model))
+
+    workgroup = min(N, 256)
+    worksize = N
+    kernel! = fill_single_buffer_variable!(dev, workgroup, worksize)
+
+    for variable in buffer_variables(particles.biogeochemistry)
+        kernel!(particles.buffer[variable], Val(variable), particles, model.grid, fields(model))
+    end
```
jagoosw:
```suggestion
end

return nothing
```

PR #340, opened 2026-03-03
Skips filling immersed cells to prevent NaNing

jagoosw:
Aiming to make it so that interior nodes don't just get filled with invalid fill value (since they always sum to zero). Of course you can always mask them after but its hard to check if stuff is running right.

I think its possible to pass the active cell map to `launch!` and not launch on those nodes instead? I imagine that would be more efficicent?


johnryantaylor (commented):
This looks good to me

johnryantaylor (approved):

PR #341, opened 2026-03-09
(0.16.13) Adds NPZD mode to LOBSTER model

jagoosw:
This introduces a `nutrients` and `detritus` for `LOBSTER` which each have only a single tracer `N` and `D` so when activated the `LOBSTER` model only has `NPZD`, or the DIC/Alk and/or oxygen can also be turned on. 

I also added the possibility to turn on temperature dependance in `PhytoZoo` with the Q10 parameterisation from the NPZD model (Kuhn 2015) which is turned on by changing the value of `temperature_coefficient` to anything other than 1 (and having a temperature tracer).

By default when these are activated the P/Z parameterisations are still from the LOBSTER model so it won't be exactly the same as the NPZD model, but its probably just a case of changing parameter values if we wanted to do that. I might remove the `NPZD` model and replace it with a function that returns this model with the parameters to replicated Kuhn 2015 NPZD.

I think this raises the question as to whether we should change the name of `LOBSTER` to something more generic (we could have a constructor that returns the LOBSTER configuration) because it now has optionally: variable C:N detritus, these options, and iron limitation, and I think its likely we might develop other nutrient cycles/alkalinity parameterisations/dual carbon chem/variable Chl phytoplankton etc. from this - @johnryantaylor 

I haven't done any validation or testing beyond checking that it steps yet but wanted to open the discussion about renaming while I worked on it.

CC: @lgloege - you could use this to have NPZD with DIC/Alk


jagoosw:
This PR takes the number of possible "LOBSTER" configurations from 32 to 72 which is going to have a big effect on testing time since most of the LOBSTER testing time is compiling the functions for all of the configurations. I've reworked them a bit (mainly for readability), but I can't think of a way that we still test all the possible configurations but don't have this big compilation expense?

I think its valuable to test all of the configurations because e.g. it caught a mistake I made with the new `Nutrient`, and I didn't know that it didn't work with the oxygen module until I ran the tests.

jagoosw:
On another note, I don't think the oxygen formulation is correct:
```julia
Rp * μP - (Rp - Rn) * nitrate_production - Rp * μNH₄
```
?

src/Models/AdvectedPopulations/LOBSTER/detritus.jl
```diff
@@ -253,5 +253,65 @@ function VariableRedfieldDetritus(grid;
...
+@kwdef struct Detritus{FT, SS}
+      remineralisation_rate :: FT = 0.1213 / day
+    small_particle_fraction :: FT = 0.5 # the fraction available to be grazed
+             redfield_ratio :: FT = 6.56 # C:N
+    
+             sinking_speeds :: SS = (u = ZeroField(), v = ZeroField(), w = ConstantField(-2.7489/day))
+end
+
+function Detritus(grid; 
+                  sinking_speed = 0.1213 / day, # m/s
```
johnryantaylor: I think this is wrong - it looks like a copy/paste error for the default remineralisation rate.

src/Models/AdvectedPopulations/LOBSTER/biology.jl
```diff
@@ -134,20 +136,42 @@ end
...
 
     Lₙ = nutrient_limitation(lobster, i, j, k, fields, auxiliary_fields)
     Lₗ = PAR / (kPAR + PAR)
+    Lₜ = Q10 ^ (T / 10)
 
-    return μ₀ * Lₗ * Lₙ * P
+    return μ₀ * Lₗ * Lₙ * Lₜ * P
+end
+
+@inline function possibly_get_T(i, j, k, fields::NamedTuple{tracer_names}) where tracer_names
```
johnryantaylor: Can we rename this function?  I find "possibly" confusing.  I also think that it would be better to throw an error (or at least a warning) if the user has set Q10 but hasn't specified the temperature.  For example, a user might have a field called "Temp" that they assume will work for temperature, but here the model will silently ignore the temperature field (by returning zero).

src/Models/AdvectedPopulations/LOBSTER/biology.jl
```diff
@@ -134,20 +136,42 @@ end
...
+    if :T ∈ tracer_names # this is a compile time evaluation
+        return fields.T[i, j, k]
```
johnryantaylor: What if temperature is stored as an auxiliary field?

src/Models/AdvectedPopulations/LOBSTER/detritus.jl
```diff
@@ -24,7 +24,7 @@ an alternative.
 """ 
 @kwdef struct TwoParticleAndDissolved{FT, SS, LS}
     remineralisation_inorganic_fraction :: FT = 0.0     # 
-            small_reminerlisation_rate :: FT = 5.88e-7 # 1/s
```
johnryantaylor, line 27: "remineralisation" is misspelled in these constants.

src/Models/AdvectedPopulations/LOBSTER/coupling_utils.jl
```diff
@@ -1,45 +1,52 @@
...
-end
-
-@inline function conserved_tracers(::LOBSTER{<:INCLUDES_NITRATE_AMMONIA, <:PhytoZoo, <:VariableRedfieldDetritus}, labeled = false)
-    nitrogen = (:NO₃, :NH₄, :P, :Z, :sPON, :bPON, :DON)
-
-    return labeled ? (; nitrogen) : nitrogen
+@inline function conserved_nitrogen_tracers(lobster::LOBSTER)
+    tracers = (:P, :Z)
+
+    if lobster.nutrients isa INCLUDES_NITRATE_AMMONIA
```
johnryantaylor: Is there a reason why this is fully capitalized?  We don't use underscores elsewhere either.  Also, do we need "Includes" here, or could we just call this "NitrateAmmonia"

johnryantaylor (commented):
@jagoosw Thanks for these edits - a nice step towards generalizing the model.  I left a few comments.

src/Models/AdvectedPopulations/LOBSTER/biology.jl
jagoosw: fields is tracers, auxiliary fields, and biogeochemical auxiliary fields. I'll change `tracer_names` to `field_names`

src/Models/AdvectedPopulations/LOBSTER/coupling_utils.jl
jagoosw: This was because `INCLUDES_NITRATE_AMMONIA` is a constant (which should be capitalised snake_case), but I realised its actually a type alias so I'll change it to CamelCase

jagoosw: Also `NitrateAmmonia` is already a type, and this is an alias to `NitrateAmmonia` and `NitrateAmmoniaIron`

src/Models/AdvectedPopulations/LOBSTER/biology.jl
jagoosw: I've changed the behaviour so the default is Q10=nothing and if its not nothing we get fields.T so an error will be thrown if they don't supply T.

src/Models/AdvectedPopulations/LOBSTER/biology.jl
jagoosw: Superseded

src/Models/AdvectedPopulations/LOBSTER/coupling_utils.jl
johnryantaylor: Ok, that makes sense

johnryantaylor (approved):
looks good to me now, thanks!

PR #354, opened 2026-04-22
Allows surface light to be functions or fields

jagoosw:
This PR finally changes the light attenuation models to allow the surface PAR to be continuous or discrete functions, or arrays, or a number, using the `getbc` interface.

### DIFF
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -129,6 +129,7 @@ function TwoBandPhotosyntheticallyActiveRadiation(; grid::AbstractGrid{FT},
                             regularize_field_boundary_conditions(
                                 FieldBoundaryConditions(top = ValueBoundaryCondition(surface_PAR)), grid, :PAR))
 
+    # wrap surface_PAR to make it work with the `getbc` interface
     surface_PAR = materialize_condition(surface_PAR, parameters, discrete_form, ()) 
     surface_PAR = regularize_boundary_condition(surface_PAR, grid, (Center(), Center(), Center()), 3, RightBoundary, nothing)
 
--- a/src/Light/2band.jl
+++ b/src/Light/2band.jl
@@ -1,11 +1,9 @@
-@kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, P, surface_PAR, t, PAR_model)
+@kernel function update_TwoBandPhotosyntheticallyActiveRadiation!(PAR, grid, clock, P, surface_PAR, PAR_model)
     i, j = @index(Global, NTuple)
 
     k, k′ = domain_boundary_indices(RightBoundary(), grid.Nz)
 
-    X = z_boundary_node(i, j, k′, grid, Center(), Center())
-
-    PAR⁰ = surface_PAR(X..., t)
+    PAR⁰ = getbc(surface_PAR, i, j, grid, clock, P)
 
     kʳ = PAR_model.water_red_attenuation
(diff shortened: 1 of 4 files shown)


src/Light/multi_band.jl
```diff
@@ -126,20 +144,18 @@ end
 @inline number_subscript(digits::NTuple{N}) where N =
     ntuple(n->Symbol(Char('\xe2\x82\x80'+digits[n])), Val(N))
 
-@kernel function update_MultiBandPhotosyntheticallyActiveRadiation!(grid, field, kʷ, e, χ,
-                                                                    _surface_PAR, surface_PAR_division, 
-                                                                    Chl, t, k′) 
+@kernel function update_MultiBandPhotosyntheticallyActiveRadiation!(grid, clock, field, kʷ, e, χ,
+                                                                    _surface_PAR, 
+                                                                    Chl, k′)
```
johnryantaylor: Why was surface_PAR_division removed here?  That variable was used to call this function in the kernel below.

src/Light/2band.jl
```diff
@@ -118,6 +131,9 @@ function TwoBandPhotosyntheticallyActiveRadiation(; grid::AbstractGrid{FT},
                             regularize_field_boundary_conditions(
                                 FieldBoundaryConditions(top = ValueBoundaryCondition(surface_PAR)), grid, :PAR))
 
+    surface_PAR = materialize_condition(surface_PAR, parameters, discrete_form, ()) 
+    surface_PAR = regularize_boundary_condition(surface_PAR, grid, (Center(), Center(), Center()), 3, RightBoundary, nothing)
+
```
johnryantaylor, line 135: It isn't going to be clear to the user what these functions do (unless they are familiar with the inner workings of Oceananigans).  Can you add a short comment to explain what is going on here?

src/Light/multi_band.jl
```diff
@@ -126,20 +144,18 @@ end
 @inline number_subscript(digits::NTuple{N}) where N =
     ntuple(n->Symbol(Char('\xe2\x82\x80'+digits[n])), Val(N))
 
-@kernel function update_MultiBandPhotosyntheticallyActiveRadiation!(grid, field, kʷ, e, χ,
-                                                                    _surface_PAR, surface_PAR_division, 
-                                                                    Chl, t, k′) 
+@kernel function update_MultiBandPhotosyntheticallyActiveRadiation!(grid, clock, field, kʷ, e, χ,
```
johnryantaylor, line 147: why was time removed as a parameter to this function?  Isn't that going to be needed in some use cases?

src/Light/multi_band.jl
```diff
@@ -157,15 +173,14 @@ function update_biogeochemical_state!(model, PAR::MultiBandPhotosyntheticallyAct
```
johnryantaylor: do we need k'?  It doesn't look like this is ever used.  As far as I can tell, the function uses grid.Nz instead.

johnryantaylor (commented):
Thanks @jagoosw.  A few comments here.

johnryantaylor (commented):
It looks like some tests are failing?

src/Light/multi_band.jl
jagoosw: `getbc` takes clock not time and the standard call order is (grid, clock, field, everyting else...) so I changed the order

johnryantaylor:
lgtm now, thanks!

PR #367, opened 2026-05-07
Secondary precipitation

johnryantaylor:
Created a simple biogeochemical model to model secondary precipitation of CaCO3.  The model includes DIC, Alk, and CaCO3 particulates. A validation script compares against shipboard experiments reported in Hashim et al. 2025.  For this, I simply folllowed the example to create a new biogeochemical model.  We should think about the best way to include this capability.


jagoosw:
Thank you for making this (and for catching the bug in the equilibrium constants, I might make a separate PR to change that right away)!

This is encouraging, typically CaCO3 is sometimes considered part of the detritus category, I think because you might want to use whatever aggregation/sinking/etc. processes that the particles experience, and as primary precipitation comes from the same processes like grazing waste. 

I'm not sure this makes sense in this kind of framework because that means we don't have the flexibility of using different detritus models. I think it would be best to have this as an optional extension of the `CarbonateSystem`, and then have that call the plankton model for primary precipitation, and the detritus model for any particle processes.

Screenshot 2026-05-07 at 16 41 49

I'm happy to make the integration if you're happy with that?

src/Models/CarbonChemistry/equilibrium_constants.jl
```diff
@@ -755,18 +755,20 @@ end
 
     pressure_correction = c.pressure_correction(T, P)
 
-    lnK_therm = c.therm_constant + c.therm_T * T + c.therm_inverse_T / T + c.therm_log_T * log10(T) # seems wrong
-    lnK_sea = ((c.sea_sqrt_S + c.sea_T_sqrt_S * T + c.sea_inverse_T_sqrt_S / T) * √S 
-                + c.sea_S * S
-                + c.sea_S_sqrt_S³ * S^convert(FT, 1.5))
-
-    return  pressure_correction * exp(lnK_therm + lnK_sea)
+    # The Mucci (1983) / Millero (2007) formula gives log₁₀(KSP), so the
```
jagoosw: Do you know where this is in the paper? I can't see any log10s?

jagoosw: Image

jagoosw: I think there is a problem with the current calcite saturation, I think it is off by a factor of 1000, but I thought currently that it might be a unit problem with the Ca+ ions?

johnryantaylor: It is in Mucci 1983 page 792.  It is confusing because it just says "log", but below equation 10, the text says that log refers to log_10.  It is worth double checking with another source, though.  Presumably CO2SYS has this formula in it?

johnryantaylor: The image that you posted does have ln().  Is this from Millero?

jagoosw: Yeah this is from Millero, I guess that this is a mistake in Millero because pyco2sys has 10^

jagoosw: I'll make a separate PR to correct this

johnryantaylor: This is what CO2SYS has: 
    % CalciteSolubility:
    % '       Mucci, Alphonso, Amer. J. of Science 283:781-799, 1983.
    logKCa = -171.9065 - 0.077993.*TempK + 2839.319./TempK;
    logKCa = logKCa + 71.595.*logTempK./log(10);
    logKCa = logKCa + (-0.77712 + 0.0028426.*TempK + 178.34./TempK).*sqrSal;
    logKCa = logKCa - 0.07711.*Sal + 0.0041249.*sqrSal.*Sal;
    % '       sd fit = .01 (for Sal part, not part independent of Sal)
    KCa = 10.^(logKCa);% ' this is in (mol/kg-SW)^2

johnryantaylor: That is the Matlab implementation of CO2SYS.

johnryantaylor: While I'm looking at it, the pressure correction term for aragonite in CO2SYS looks very different:
    % PressureCorrectionForAragonite:
    % '       Millero, Geochemica et Cosmochemica Acta 43:1651-1661, 1979,
    % '       same as Millero, GCA 1995 except for typos (-.5304, -.3692,
    % '       and 10^3 for Kappa factor)
    deltaVKAr = deltaVKCa + 2.8;
    KappaKAr  = KappaKCa;
    lnKArfac  = (-deltaVKAr + 0.5.*KappaKAr.*Pbar).*Pbar./RT;
    KAr       = KAr.*exp(lnKArfac);

The pressure correction for calcite is the same as what we have.  Is it possible that this needs to be updated too?

johnryantaylor: Sorry, scratch that last comment -  I hadn't spotted that CO2SYS is just modifying the calcite values. This looks fine

johnryantaylor:
@jagoosw , just to add that I'm happy for you to integrate this into CarbonateSystem.  Let me know if I can help.

src/Models/CarbonChemistry/equilibrium_constants.jl
jagoosw: I found how this slipped through, the book I used for the check values also just lists "log", and I realise now they use "ln" for the natural log
Screenshot 2026-05-08 at 17 26 04

johnryantaylor:
@jagoosw, I added an example script which demonstrates particle sinking and dissolution.

PR #362, opened 2026-04-30
Generalises carbon chemistry to allow multiple instances

jagoosw:
This PR generalises `CarbonateSystem` to allow `N` instances of DIC and alkalinity by passing the number as an argument e.g. `CarbonateSystem(2)`. This makes N copies of DIC and Alk named DIC1, DIC2, ..., DICN etc.

This PR also adds an idealised OAE example.

total_change

https://github.com/user-attachments/assets/b443c8ef-3f8e-4231-88b9-3030df2c9c51

To do:

- [x] Check example
- [x] Add tests for `CarbonDioxideConcentration` and `CarbonateSystem`
- [x] Update docstrings

@ElizabethYankovsky @lgloege @johnryantaylor 

~This PR currently includes #355 but I will rebase when it is merged.~ done

~Blocked by #355,~ closes #363 and #361

PR #370, opened 2026-05-08
Generalises gas transfer velocity and allows `surface_value` to interpolate `FieldTimeSeries`

### LINKED ISSUE #368: Generalise gas transfer velocity
@lgloege

jagoosw:
Closes #368 @lgloege

I'm a little concerned about how deep the parameter space is for these boundary conditions and how that might effect peformance.


src/Models/GasExchange/carbon_dioxide_concentration.jl
jagoosw, line 1: This is from #362 don't know how it got into this PR

jagoosw:
@lgloege As an example of how this could now be generalised, if you wanted a transfer velocity dependant on wind speed, temperature, and significant wave height, $k(u, T, H)$, you could make a struct like:

```julia
struct WindWaveTempTransferVelocity{FT, WS, WH}
    a_constant :: FT
    wind_speed :: WS
   wave_height :: WH
end
```

and define a function:
```julia
using OceanBioME.Models.GasExchangeModel: surface_value

@inline function (k::WindWaveTempTransferVelocity)(i, j, grid, clock, model_fields)
    T = @inbounds model_fields.T[i, j, grid.Nz]
    u = surface_value(k.wind_speed, i, j, grid, clock, model_fields)
    H = surface_value(k.wave_height, i, j, grid, clock, model_fields)

    return k.a_constant * u * H
end
```

Then you can make an instance like:
```julia
k0 = WindWaveTempTransferVelocity(0.1, wind_speed, wave_height)
```
where `wind_speed` and `wave_height` are number, arrays (in space not time), fields, or field time series (on the same grid as the model or not). When you actually make the boundary condition you do need to wrap this in the Schmidt number scaling:
```julia
k = SchmidtScaledTransferVelocity(
         schmidt_number = CarbonDioxidePolynomialSchmidtNumber(), 
         solubility = MolPerKgPerAtmToMMolPerCubicMPerMicroAtm(carbon_chemistry.solubility, carbon_chemistry.density_function),
         base_transfer_velocity = k0
     )
```
which goes in the boundary condition constructor:
```julia
CarbonDioxideGasExchangeBoundaryCondition(; transfer_velocity = k)
```

This isn't the best interface but we can work on that.

Perhaps there is a way NumericalEarths atmosphere might carry around the waves (or someone might write a wave model in Julia that could be coupled), but for now it seems most straight forward to directly put the wave info in this way.

PR #373, opened 2026-05-13
[WIP] adds ammonia dissociation to alkalinity

jagoosw:
This PR explores adding ammonia dissociation:
```math
NH_4^+ 
\rightleftharpoons NH_3 + H^+
```
to the pH computation where $[NH_3]$ contributes to the total alkalinity. The equilibrium is parameterised as:
```math
[NH_3] = \frac{[NH_{3T}]}{1+[H]/K}
```
where  $[NH_{3T}] = [NH_3] + [NH_4^+]$ which (I think) is what we model as $NH_4$.

Lots of models neglect this contribution and I thought it might have en effect abnormal since $K\approx2.5\times10^{-10}$ at T = 35C, S = 35PSU, so at $pH=10$ $[NH_3] = [NH_{3T}]$. 

I'm not sure what to make of the results. Here is the difference pCO2 when accounting for the ammonia contribution or not at "normal" alkalinity ($2100meq/m^3$) us blue and high alkalinity ($3000$).
image
Here instead we vary alkalinity and plot the difference between accounting for $50mmol/m^3$ of ammonia vs not:
image

or plot the pH
image

And the dissotiation as expected:
image

(all plots at 15C, 35PSU and $DIC = 2070mmol/m^3$ where relivant)

Cons: I think most models assume that the contribution is small enough to not effect the results to known precision. Also, this increases the cost of the solve by slightly more than 2x, maybe this is curable by adjusting the preconditioning but possibly not.

Pros: this is maybe "more correct" since we normally know $[NO_{3T}]$ but this is very hard to verify since lots of data doesn't include ammonia. When we include this infrastructure but don't use it, the solve is as fast as it is without (when I inspect the compiled code I think it compiles the zeros away).

Perhaps it would be good todo some analysis of what we actually need to include because e.g. we carry around phosphate and silicate contributions which we have never used.

In addition to this PR, if we want to include this contribution, or that of phosphate or silicate we need to modify this function to optionally pass the relevant concentrations:
```julia
@inline function surface_value(cc::CarbonDioxideConcentration{DIC_name, Alk_name}, i, j, grid, clock, model_fields) where {DIC_name, Alk_name}
    DIC = @inbounds model_fields[DIC_name][i, j, grid.Nz] # this is a compile time inference so is fine on GPU
    Alk = @inbounds model_fields[Alk_name][i, j, grid.Nz]

    T = @inbounds model_fields.T[i, j, grid.Nz]
    S = @inbounds model_fields.S[i, j, grid.Nz]

    silicate, phosphate = silicate_and_phosphate(cc.silicate_and_phosphate_names, model_fields)

    pCO₂ = cc.carbon_chemistry(; DIC, Alk, T, S, silicate, phosphate, output = Val(:pCO₂))

    return pCO₂ # ppmv
end
```
Probably we can do the `:Si in keys(model_fields)` which gets evaluated at compile time trick.


jagoosw:
Closing as I don't think this is worthwhile at the moment

PR #377, opened 2026-05-21
(0.18.1) Adds explicit PIC

jagoosw:
Following #367 this PR adds an explicit calcite DIC/Alk/CaCO3 model which goes in the `inorganic_carbon` slot.

By default no abiotic precipitation occurs. For dissolution and precipitation parameters you can specify a number or a tuple to use different numbers for each replicate.

I'm going to update the oae example to use this with three replicates: control, release with no precipitation, release with precipitation

@johnryantaylor @lgloege


jagoosw:
There are now 384 combinations of nutrients/plankton/detritus/inorganic carbon/oxygen being tested so we might need to reconsider how they're tested

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/carbonate_system.jl
```diff
@@ -34,44 +48,253 @@ integer argument when the model is constructed like `CarbonateSystem(2)`.
...
+                            adapt(to, cs.precipitation_rate_constant),
+                            adapt(to, cs.precipitation_order),
+                            adapt(to, cs.calcite_sinking_velocity))
+
+function update_biogeochemical_state!(model, bgc::CarbonateSystem{N, true}) where N
+    carbon_chemistry = bgc.carbon_chemistry
+    grid = model.grid
+
+    T = model.tracers.T
+    S = model.tracers.S
```
johnryantaylor: explicit_calcite=true requires T and S to be model variables, but this isn't clear to the user

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/detritus.jl
```diff
@@ -90,7 +90,6 @@ required_biogeochemical_tracers(::VariableRedfieldDetritus) = (:sPOC, :bPOC, :DO
 
 @inline (bgc::NutrientsPlanktonDetritus{<:Any, <:Any, <:VariableRedfieldDetritus})(i, j, k, grid, ::Val{:bPOC}, clock, fields, auxiliary_fields) = (
     (1 - bgc.detritus.small_solid_waste_fraction) * solid_carbon_waste(bgc, i, j, k, fields, auxiliary_fields)
-  + calcite_production(bgc, i, j, k, fields, auxiliary_fields)
```
johnryantaylor: what happens if we aren't using explicit calcite?

johnryantaylor:
Sorry to be a pain, but I wonder if we should change "calcite" to PIC so that the naming convention is more generic.  In some OAE contexts we might want this to be aragonite which has a different solubility compared to calcite.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/carbonate_system.jl
johnryantaylor: I added T and S as required tracers when explicit_calcite=true

johnryantaylor:
> Sorry to be a pain, but I wonder if we should change "calcite" to PIC so that the naming convention is more generic. In some OAE contexts we might want this to be aragonite which has a different solubility compared to calcite.

I'm still not sure the answer to this.  Do we want to leave this as calcite and add functionality for aragonite if needed, or have a generic form of precipitated mineral for which we can change the constants if needed?  @jagoosw do you have any thoughts on this?

jagoosw:
> > Sorry to be a pain, but I wonder if we should change "calcite" to PIC so that the naming convention is more generic. In some OAE contexts we might want this to be aragonite which has a different solubility compared to calcite.
> 
> I'm still not sure the answer to this. Do we want to leave this as calcite and add functionality for aragonite if needed, or have a generic form of precipitated mineral for which we can change the constants if needed? @jagoosw do you have any thoughts on this?

I think we can rename it to PIC, but I think it might make sense todo some more refactoring of this PR after #378

jagoosw:
@johnryantaylor following #378 I have updated this PR

johnryantaylor:
> > > Sorry to be a pain, but I wonder if we should change "calcite" to PIC so that the naming convention is more generic. In some OAE contexts we might want this to be aragonite which has a different solubility compared to calcite.
> > 
> > 
> > I'm still not sure the answer to this. Do we want to leave this as calcite and add functionality for aragonite if needed, or have a generic form of precipitated mineral for which we can change the constants if needed? @jagoosw do you have any thoughts on this?
> 
> I think we can rename it to PIC, but I think it might make sense todo some more refactoring of this PR after #378

@jagoosw, is now a good time to revisit the naming convention?  Here are some new thoughts: I think that ExplicitCalciumCarbonate would be a good name for this model.  That is more generic than calcite, and the user can set the appropriate saturation state depending which polymorph is appropriate for their application.  For the variable names, I suggest CaCO_3.  This is what MARBL uses.   When naming functions and parameters, I suggest that we use "calcium_carbonate" where "calcite" is currently used.  I think I prefer that to "PIC" and "particulate_inorganic_carbon".  If that sounds ok, what is the best way to make that change?  I'm happy to make those changes, but it will be a lot of edits!

jagoosw:
> @jagoosw, is now a good time to revisit the naming convention? Here are some new thoughts: I think that ExplicitCalciumCarbonate would be a good name for this model. That is more generic than calcite, and the user can set the appropriate saturation state depending which polymorph is appropriate for their application. For the variable names, I suggest CaCO_3. This is what MARBL uses. When naming functions and parameters, I suggest that we use "calcium_carbonate" where "calcite" is currently used. I think I prefer that to "PIC" and "particulate_inorganic_carbon". If that sounds ok, what is the best way to make that change? I'm happy to make those changes, but it will be a lot of edits!

Yes that seems like a good idea, I'm happy with renaming everything from calcite to calcium_carbonate. I'll try and get it done tomorrow morning.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/utils.jl
```diff
@@ -61,33 +61,54 @@ for thing in (PhytoZoo, Detritus, DissolvedParticulate)
...
 
 group_element_tracers(::CarbonateSystem, args...) = NamedTuple()
-group_element_tracers(::CarbonateSystem, bgc::NPD{FT}, ::Val{:carbon}) where FT = 
+group_element_tracers(::CarbonateSystem, bgc::NPD{FT}, ::Val{:carbon}) where FT =
     (; DIC = one(FT))
 
+group_element_tracers(::ExplicitCalcite, args...) = NamedTuple()
+group_element_tracers(::ExplicitCalcite{1}, bgc::NPD{FT}, ::Val{:carbon}) where FT =
+    (DIC = one(FT), CaCO₃ = one(FT))
+function group_element_tracers(ic::ExplicitCalcite{N}, bgc::NPD{FT}, ::Val{:carbon}) where {N, FT}
```
jagoosw: I think this is wrong

jagoosw:
I'm not sure why the alkalinity tests are failing now they seem fine in main

PR #365, opened 2026-05-04
Fixes alkalinity formulaiton

jagoosw:
Closes #364 

To do:
- [x] Update documentation

PR #378, opened 2026-05-22
0.18.0 - Major internal refactor of `NutrientsPlanktonDetritus` abstraction + adds `ImplicitProductivity` plankton

jagoosw:
This PR has turned into a refactor of the NPD module which hopefully aids abstraction. It also adds an `ImplicitProductivity` `plankton` (details below). 

I will update docs (probably the "Implementing new models page") but this refactor hopefully aids implementing new elements by specifying the basic interfaces between `nutrients`, `plankton`, `detritus`, `inorganic_carbon`, and `oxygen`, but more complex models can still write their own methods up to `(bgc::NutrientPlanktonDetritus{FT, <:Nutrients{<:Any, <:MyIronCycle}, <:MyNewPlankton)(i, j, k...)`.

This means that if you want to implement a new thing, for example a plankton, you just define the nutrient usage, elemental ratios, and waste (inorganic, dissolved, solid), and if the new model is not too abnormal it should just work. If you e.g. wanted to have a phytoplankton with variable C:N you would then need to write methods for `inorganic_nitrogen_waste` and `inorganic_carbon_waste` etc. but it would still work if you e.g. included phosphate growth limitation with no additional work.

The implicit `plankton` which tracks no tracers but computes a community productivity like [this model from MITGCM](https://mitgcm.readthedocs.io/en/latest/examples/global_oce_biogeo/global_oce_biogeo.html#sub-global-oce-biogeo) and [ClimaOceanBiogeochemistry](https://github.com/CliMA/ClimaOceanBiogeochemistry.jl/blob/main/src/carbon_alkalinity_nutrients.jl).

This might be useful as a starting point for global runs so we can see if this simple model replicates other results (vs e.g. not being able to know if biases are from LOBTER or from the circulation, since there are no global LOBSTER solutions).

To do:
- [x] check the examples work
- [x] doc strings
- [x] write docs page about the abstraction
- [x] docs page about implicit productivity

Breaking changes:
- Moves `grid` and `surface_PAR` to be positional arguments of light models
- Changes `carbonate_system` to `inorganic_carbon`
- `VariableRedfieldDetritus` is now called `CarbonNitrogenDissolvedParticulate`


jagoosw:
@johnryantaylor this PR maintains almost all existing functionality, except variable C:N detritus. I'm still thinking about the best way to abstract this and as its our only thing that has variable elemental ratios I think its worth considering how best to do it for e.g. when we implement MARBL plankton. 

I also think that once we have devised this interface, it would be straight forward to have one of the "elements" of the large particles be calcite (or multiple calcium carbonate species/different precipitation methods), because `calcite_dissolution` is already a property of the `detritus` (and is otherwise where particulate detritus goes).

jagoosw:
@johnryantaylor @nanophyto @lgloege

Something I've been thinking about is where to land on the balance of code complexity vs writing out similar things for different situations. As an example I implemented `DissolvedParticulate` to allow any number of dissolved and/or particulate classes for e.g. size or lability, which obfuscates the code to some extent because the methods gets wrapped up in expressions: 

https://github.com/OceanBioME/OceanBioME.jl/blob/13cdab61fb94d28eba8bf8141eff9f4613b219a7/src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl#L92-L178

I think in this case it is probably still clear enough to land on the right side of the complexity/obfuscation balance, but I've been trying to work out how to extend this for variable elemental ratio detritus and I'm not sure if it becomes too difficult to understand the code? 

It gets relatively confusing because if we explicitly have the default plankton element represented (e.g. nitrogen or phosphate) only, then we have to assume that the particulates contain carbon and calcite in the same ratio as the plankton (and that the zooplankton are the same as the phytoplankton), but if we explicitly represent carbon then we can associate that calcite with the POC groups so the zooplankton doesn't have to have calcite associated with it (in LOBSTER grazing leads to some dissolution and some PIC production accounting for all the calcite in the phytoplankton). But we could also have explicit calcite groups which means the POC group now only has the organic carbon. This can be even more complicated if we then want multiple DIC/Alk because we have to have different PIC groups then because they'll dissolve differently.

(Having different DIC/Alk replicates could actually mean you need to have replicates of the plankton in some models, for example PISCES has the rain ratio in the phytoplankton vary depending on saturation state. MARBL doesn't do that and only varies rain ratio by other environmental conditions. Perhaps its not a first order effect but in the context of OAE it might actually be something we would want to capture, but would make the modelling significantly harder.)

I was wondering what you all thought of where on the code complexity/repetition we should land and if we're going to invest in more complex code that automates model setup how we should allow for future development of new elements e.g. aggregation? Perhaps there is a better abstraction for detritus that could solve some of these problems.

@glwagner do you have any thoughts on this too?

glwagner:
> @johnryantaylor @nanophyto @lgloege
> 
> Something I've been thinking about is where to land on the balance of code complexity vs writing out similar things for different situations. As an example I implemented `DissolvedParticulate` to allow any number of dissolved and/or particulate classes for e.g. size or lability, which obfuscates the code to some extent because the methods gets wrapped up in expressions:
> 
> https://github.com/OceanBioME/OceanBioME.jl/blob/13cdab61fb94d28eba8bf8141eff9f4613b219a7/src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl#L92-L178
> 
> I think in this case it is probably still clear enough to land on the right side of the complexity/obfuscation balance, but I've been trying to work out how to extend this for variable elemental ratio detritus and I'm not sure if it becomes too difficult to understand the code?
> 
> It gets relatively confusing because if we explicitly have the default plankton element represented (e.g. nitrogen or phosphate) only, then we have to assume that the particulates contain carbon and calcite in the same ratio as the plankton (and that the zooplankton are the same as the phytoplankton), but if we explicitly represent carbon then we can associate that calcite with the POC groups so the zooplankton doesn't have to have calcite associated with it (in LOBSTER grazing leads to some dissolution and some PIC production accounting for all the calcite in the phytoplankton). But we could also have explicit calcite groups which means the POC group now only has the organic carbon. This can be even more complicated if we then want multiple DIC/Alk because we have to have different PIC groups then because they'll dissolve differently.
> 
> (Having different DIC/Alk replicates could actually mean you need to have replicates of the plankton in some models, for example PISCES has the rain ratio in the phytoplankton vary depending on saturation state. MARBL doesn't do that and only varies rain ratio by other environmental conditions. Perhaps its not a first order effect but in the context of OAE it might actually be something we would want to capture, but would make the modelling significantly harder.)
> 
> I was wondering what you all thought of where on the code complexity/repetition we should land and if we're going to invest in more complex code that automates model setup how we should allow for future development of new elements e.g. aggregation? Perhaps there is a better abstraction for detritus that could solve some of these problems.
> 
> @glwagner do you have any thoughts on this too?

I think in general you have to take an empirical approach here... sometimes it can be better to error on the side of clarity (ie duplicating some code) at first, and then eliminate duplications if maintenance becomes hard, later.

Would it help if you develop an interface for the "tracer name type" in Oceananigans? It seems like at least some of the difficulty comes from the fact that you need to dispatch on `::Val{name}`. Not sure that would help.

I noticed that you have functions like

```julia
dissolved_waste(bgc.plankton, bgc, i, j, k, fields, auxiliary_fields)
```

I recommend always using the "kernel form" `dissolved_waste(i, j, k, grid, args...)`. Otherwise, users cannot write diagnostics (without a wrapper) which evaluates that RHS term. If you use the kernel form, you can call the functions from `KernelFunctionOperation`.

jagoosw:
> I think in general you have to take an empirical approach here... sometimes it can be better to error on the side of clarity (ie duplicating some code) at first, and then eliminate duplications if maintenance becomes hard, later.
> 
> Would it help if you develop an interface for the "tracer name type" in Oceananigans? It seems like at least some of the difficulty comes from the fact that you need to dispatch on `::Val{name}`. Not sure that would help.
> 

This might be helpful because some of the expressions obfuscation comes from having to put names in. What sort of solution do you think it could look like? 

> I noticed that you have functions like
> 
> ```julia
> dissolved_waste(bgc.plankton, bgc, i, j, k, fields, auxiliary_fields)
> ```
> 
> I recommend always using the "kernel form" `dissolved_waste(i, j, k, grid, args...)`. Otherwise, users cannot write diagnostics (without a wrapper) which evaluates that RHS term. If you use the kernel form, you can call the functions from `KernelFunctionOperation`.

Thats a good point, I always went with this order because of some notion that you should have arguments you're dispatching on first, but I think I made this up and it would be beneficial to be able to use KFOs.

I was also wondering how complex the micro physics was getting in Breeze and if you'd had any thoughts on how handle model choices flexibly (e.g. I gather you might want to track different size classes of precipitation sometimes)?

johnryantaylor:
Thanks for this @jagoosw!  I really like the concept which could let the user build models of varying complexity.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl
```diff
@@ -0,0 +1,237 @@
...
+default_sinking_speeds(::Symbol, FT=Float64) = convert(FT, 10/day)
+
+default_partitioning(names, FT=Float64) = tuple(repeat([convert(FT, 1/length(names))], length(names))...)
+default_partitioning(::Symbol, FT=Float64) = one(FT)
+
+function manifest_multi_class_dissolved_particulate(dissolved_names, particulate_names)
+    dissolved_names = possibly_tuple_or_symbol(dissolved_names)
+    particulate_names = possibly_tuple_or_symbol(particulate_names)
+
+    for (n, name) in enumerate(dissolved_names)
```
johnryantaylor, line 126: Naive question: how is this going to dispatch if there are multiple  models with the same number of elements?

jagoosw: If the new model has particles with the same names it will overwrite the method which gives a warning but is otherwise fine because its doing the same thing for every model just with different parameters.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    for (n, name) in enumerate(dissolved_names)
+        ex = :(total += fields[$(QuoteNode(name))][i, j, k] * 
+                        detritus.dissolved_remineralisation_rate[$n])
+        push!(ex_inorganic_remineralisation.args, ex)
+    end
+
+    @eval begin
+        @inline function dissolved_remineralisation(i, j, k, grid, detritus::DissolvedParticulate{$N, $M}, bgc::NPD_DP{FT}, fields, auxiliary_fields) where FT
```
johnryantaylor: These three functions don't dispatch on the variable names.  I think this will cause a conflict when creating two bgc models with the same number of components but with different names.

johnryantaylor: @jagoosw, I have confirmed that this is a problem.  This code works:
```
bgc_a = NutrientsPlanktonDetritus(grid;
            nutrients = Nutrients(nitrogen = NitrateAmmonia()),
            plankton = PhytoZoo(grid),
            detritus = DissolvedParticulate(grid, :DOM, (:sPOM, :bPOM)),
            light_attenuation = par)
model_a = NonhydrostaticModel(grid; biogeochemistry = bgc_a, advection = nothing)
```
but the following throws an error because the creation of bgc_b overwrites bca_a since dissolved particulate dispatches on the number of variables but not the variable name:
```
bgc_a = NutrientsPlanktonDetritus(grid;
            nutrients = Nutrients(nitrogen = NitrateAmmonia()),
            plankton = PhytoZoo(grid),
            detritus = DissolvedParticulate(grid, :DOM, (:sPOM, :bPOM)),
            light_attenuation = par)
bgc_b = NutrientsPlanktonDetritus(grid;
            nutrients = Nutrients(nitrogen = NitrateAmmonia()),
            plankton = PhytoZoo(grid),
            detritus = DissolvedParticulate(grid, :DON, (:sPON, :bPON)),
            light_attenuation = par)
model_a = NonhydrostaticModel(grid; biogeochemistry = bgc_a, advection = nothing)
```

johnryantaylor: I pushed an edit that I think fixes this (just adding the tracer names to the calling sequence for dispatch).

johnryantaylor (approved):

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl
```diff
@@ -0,0 +1,227 @@
...
+end
+
+@inline @generated function calcite_dissolution(i, j, k, grid, detritus::DissolvedParticulate{N, M, DN, PN}, bgc::NPD_DP{FT}, fields, auxiliary_fields) where {N, M, DN, PN, FT}
+    combined = Expr(:block)
+    push!(combined.args, :(total = zero($FT)))
+    for (m, name) in enumerate(PN)
+        push!(combined.args, :(total += fields[$(QuoteNode(name))][i, j, k] *
+                        detritus.particulate_remineralisation_rate[$m]))
+    end
+    push!(combined.args, :(total += dissolved_waste(i, j, k, grid, bgc.plankton, bgc, fields, auxiliary_fields)))
```
johnryantaylor, line 188: I think I'm missing something here.  Why does dissolved waste go into calcite dissolution?

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/InorganicCarbon/abstract_inorganic_carbon.jl
```diff
@@ -0,0 +1,20 @@
...
+  + inorganic_carbon_waste(i, j, k, grid, bgc.plankton, bgc, fields, auxiliary_fields)
+  + inorganic_carbon_waste(i, j, k, grid, bgc.detritus, bgc, fields, auxiliary_fields)
+  - net_calcite_production(i, j, k, grid, bgc, fields, auxiliary_fields)
+)
+
+@inline function (bgc::NPD_AIC)(i, j, k, grid, ::Val{:Alk}, clock, fields, auxiliary_fields)
+    PN = phosphate_ratio(i, j, k, grid, bgc.plankton, bgc, fields) / nitrogen_ratio(i, j, k, grid, bgc.plankton, bgc, fields)
+
+    return (
+        nitrate_primary_production(i, j, k, grid, bgc.plankton, bgc, clock, fields, auxiliary_fields) * (1 - PN)
```
johnryantaylor: Should there be a - sign in front of this line?  uptake of nitrate should increase the alkalinity.  In the nitrate_primary_production function, nutrient_uptake is negative.  Hence, without a minus sign here, it looks like nitrate uptake is reducing alkalinity instead of increasing it.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Plankton/phyto_zoo.jl
```diff
@@ -348,8 +348,8 @@ end
 @inline edible_particulate_organic_matter(i, j, k, grid, ::DissolvedParticulate{<:Any, 1, <:Any, PN}, plankton::PhytoZoo, bgc, fields) where PN = 
     @inbounds plankton.edible_fraction_of_detritus * getproperty(fields, PN[1])[i, j, k]
 
-@inline edible_particulate_organic_matter(i, j, k, grid, ::DissolvedParticulate{<:Any, 2}, plankton::PhytoZoo, bgc, fields) = 
-    @inbounds fields.sPOM[i, j, k] # this isn't generic because the user might rename the particles...
+@inline edible_particulate_organic_matter(i, j, k, grid, ::DissolvedParticulate{<:Any, 2, <:Any, PN}, plankton::PhytoZoo, bgc, fields) where PN =
```
jagoosw: @johnryantaylor this changes the behaviour of the model and is inconsistent with the LOBSTER formulation

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl
jagoosw: Okay I understand the problem now.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Detritus/single_element.jl
jagoosw: All phytoplankton waste has calcite associated with it and it is commonly assumed that calcite associated with the dissolved fraction of the waste is dissolved.

Perhaps this is biologically true or not but if you don't do this then the particulate waste has an unknowable ratio of calcite associated with it, so you have todo this at least if you're not explicitly tracking calcite.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/InorganicCarbon/abstract_inorganic_carbon.jl
jagoosw: I think this is correct: https://github.com/OceanBioME/OceanBioME.jl/pull/365#issuecomment-4440318065

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Plankton/phyto_zoo.jl
```diff
@@ -0,0 +1,446 @@
...
+    zero(FT)
+    
+@inline edible_particulate_organic_matter(i, j, k, grid, ::Detritus, plankton::PhytoZoo, bgc, fields) = 
+    @inbounds plankton.edible_fraction_of_detritus * fields.D[i, j, k]
+
+@inline edible_particulate_organic_matter(i, j, k, grid, ::DissolvedParticulate{<:Any, 1, <:Any, PN}, plankton::PhytoZoo, bgc, fields) where PN = 
+    @inbounds plankton.edible_fraction_of_detritus * getproperty(fields, PN[1])[i, j, k]
+
+@inline edible_particulate_organic_matter(i, j, k, grid, ::DissolvedParticulate{<:Any, 2, <:Any, PN}, plankton::PhytoZoo, bgc, fields) where PN =
+    @inbounds plankton.edible_fraction_of_detritus * getproperty(fields, PN[1])[i, j, k]
```
jagoosw: @johnryantaylor for things like this could you suggest edits like:

```suggestion
    @inbounds fields.sPOM[i, j, k]
```
So that we can discuss them before committing.

jagoosw: Maybe we could also do:

```suggestion
    @inbounds getproperty(fields, PN[1])[i, j, k]
```

but that means we're forcing the assumption that `PN[1]` is the smallest and edible particulate

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Plankton/phyto_zoo.jl
johnryantaylor: oh, ok, I thought that this could be made more generic.  Ok to revert back.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Plankton/phyto_zoo.jl
johnryantaylor: Good point.  Maybe we should keep it sPOM for now.  We can think about whether we want to make the naming more generic (as per your inline comment).

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/InorganicCarbon/abstract_inorganic_carbon.jl
johnryantaylor: what about the minus sign here on line 17 of defaults.jl: 
          - nutrient_uptake(i, j, k, grid, bgc.plankton, bgc, fields, auxiliary_fields)
I was concerned that nitrate uptake is decreasing alkalinity.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/Plankton/phyto_zoo.jl
jagoosw: I figured out a way around this now, sPOM is the fallback but user can specify another name

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/InorganicCarbon/abstract_inorganic_carbon.jl
jagoosw: Oh I understand now, I think I've switched the nitrate and ammonia places

johnryantaylor:
Hi @jagoosw, what are we still missing for this PR?  Do you want me to work on the docs or an example?

jagoosw:
> Hi @jagoosw, what are we still missing for this PR? Do you want me to work on the docs or an example?

I'll go through and fix the docs etc in the next few days

PR #386, opened 2026-07-18
Fixes CI problems 

jagoosw:


johnryantaylor:
@jagoosw, if this is proving to be a pain (and particularly if it relates to kelp) then I wonder if we should consider running the tests through github actions instead of buildkite.

jagoosw:
I think I've fixed the kelp related problems now, but I'm not sure why the alkalinity tests are failing now

johnryantaylor:
@jagoosw: This is the error that I see in the buildkite log from the Alkalinity test: 

  | Alkalinity sign: Test Failed at /var/lib/buildkite-agent/Oceananigans.jl-1513/test/test_NutrientsPlanktonDetritus.jl:177
  | Expression: isapprox(ΔAlk, -((1 + 1 / 16)) * ΔNO₃ + (1 - 1 / 16) * ΔNH₄)
  | Evaluated: isapprox(8.895109431250603e-6, 8.895109568287718e-6)
  
The two values are only different by about 1e-13.  That is close to the machine precision level default error of isapprox.  We could reduce the precision threshold for isapprox to see if the problem is simply machine precision.  I'll suggest a fix.

johnryantaylor:
We could change this
test/test_NutrientsPlanktonDetritus.jl:177:
    @test isapprox(ΔAlk, -(1 + 1/16) * ΔNO₃ + (1 - 1/16) * ΔNH₄)
to
    @test isapprox(ΔAlk, -(1 + 1/16) * ΔNO₃ + (1 - 1/16) * ΔNH₄; atol = 1e-10)

which should easily pass.  It looks like the default for isapprox is a relative error of sqrt(eps()) where eps()=1e-16 is machine precision for a double precision float.  I think isapprox is failing because ΔAlk and the other number are already quite small and hence it fails with a relative error of 1e-8.  Using the default of isapprox without specifying the error level that we care about might also be susceptible to failing if the default changes in different version of Julia (I'm not sure if it did change, just a guess as to why it is failing now).

Of course, it is possible that the error will build up and the test will fail at a higher level anyway if something is genuinely wrong...

jagoosw:
It was passing with the default tolerance before so I'm not sure why it changed, but we could just go with a lower tolerance

PR #391, opened 2026-08-18
(0.18.3) Makes light models more generic

nanophyto:
Generalizes `TwoBandPhotosyntheticallyActiveRadiation` so that chlorophyll attenuation is obtained from the biogeochemistry through the existing `chlorophyll(...)` interface rather than assuming a single phytoplankton tracer named `P`.

Previously, TwoBand calculated chlorophyll internally as a fixed chlorophyll-to-phytoplankton ratio multiplied by `model.tracers.P`. This tied the light model to biogeochemical models with exactly one phytoplankton tracer using that name.

With this change, TwoBand instead uses:

```julia
chlorophyll(model.biogeochemistry, model)
```

This makes the light model independent of the number, names, and representation of phytoplankton tracers.

Standard NutrientsPlanktonDetritus/LOBSTER behaviour should remain unchanged, since PhytoZoo already defines:

```julia
chlorophyll(plankton::PhytoZoo, model) =
    plankton.chlorophyll_ratio * model.tracers.P
```

The change does remove the separate `phytoplankton_chlorophyll_ratio` option from `TwoBandPhotosyntheticallyActiveRadiation`. This previously allowed TwoBand itself to convert P directly to chlorophyll. However, this appears to be redundant with the current chlorophyll(...) interface since NPD/LOBSTER now owns its chlorophyll ratio through PhytoZoo, while PISCES provides chlorophyll from it's own state representation. Nonetheless, worth checking if dropping this causes any issue? 

related to https://github.com/orgs/OceanBioME/discussions/390

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "OceanBioME"
 uuid = "a49af516-9db8-4be4-be45-1dad61c5a376"
-version = "0.18.2"
+version = "0.18.3"
 authors = ["Jago Strong-Wright <js2430@damtp.cam.ac.uk> and contributors"]
 
 [deps]
--- a/docs/src/model_components/biogeochemical/LOBSTER.md
+++ b/docs/src/model_components/biogeochemical/LOBSTER.md
@@ -146,7 +146,7 @@ Additionally, the ``DIC`` and ``Alk`` equations are modified to replace each ``X
 | ``\alpha_d``       | `ammonia_fraction_of_detritus`         | -                  |
 | ``R_P``            | `phytoplankton_redfield`               | mmol C / mmol N    |
 | ``R_O``            | `organic_redfield`                     | mmol C / mmol N    |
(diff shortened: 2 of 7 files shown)


jagoosw:
Somethings happened to the docs, I fixed it in a different PR but I'll merge just the fix to main asap so we can check the docs here

nanophyto:
Yes, I noticed. Strange. I'll keep an eye on main and merge the relevant fix when implemented (for #393 as well).

jagoosw:
Github was having some problems at some point that was [effecting code when you merge stuff etc](https://github.com/orgs/community/discussions/193645). so maybe its something like that because I dont think anyone has touched the bib file for some time.

jagoosw (approved):
LGTM when the tests pass. Thanks for sorting this!

nanophyto:
> LGTM when the tests pass. Thanks for sorting this!

Thanks Jago. It looks like the tests now pass so I will merge it.

PR #393, opened 2026-08-18
Generalise external plankton support in NutrientsPlanktonDetritus

nanophyto:
Formalises the existing plankton-component extension interface in `NutrientsPlanktonDetritus` so external packages can provide arbitrary plankton communities without depending on `PhytoZoo`-specific assumptions.

Most of the required extension points already existed so the PR is rather minimal.

I changed:

* Generalise plankton conservation metadata through the existing `group_element_tracers` dispatch.

  * Community-wide elemental ratios remain the default.
  * Individual plankton tracers can define different elemental compositions where required.
* Add community-wide and tracer-specific `chlorophyll_ratio` dispatch.
* Correct documentation referring to a separate `detritus_grazing` interface; detritus consumption already uses the generic `grazing` path.

To formalize the external-plankton coupling I also added an explicit test with:

* arbitrary external plankton tracer names;
* external plankton tendency dispatch;
* nutrient and waste exchange hooks;
* consumption of an NPD detritus tracer through `grazing`;
* community-wide and tracer-specific chlorophyll ratios;
* inclusion of external plankton tracers in elemental conservation metadata.

I opted to include the test in it's own file rather than ""test_NutrientsPlanktonDetritus.jl" since the latter takes a long time to run. But happy to move it if preferred. 

related to https://github.com/orgs/OceanBioME/discussions/390


src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/utils.jl
```diff
@@ -48,6 +47,30 @@ group_element_tracers(::ImplicitProductivity, args...) = NamedTuple()
 group_element_tracers(::InstantRemineralisationDetritus, args...) = NamedTuple()
 group_element_tracers(::Nothing, args...) = NamedTuple()
 
+function group_element_tracers(group, bgc, val_element)
+    group === bgc.plankton || return NamedTuple()
```
jagoosw: I'm not sure I understand what this method is doing? And why it would return nothing if its called on the plankton

test/test_external_plankton_interface.jl
```diff
@@ -0,0 +1,79 @@
...
+
+using OceanBioME: conserved_tracers
+using OceanBioME.Models.NutrientsPlanktonDetritusModels: InstantRemineralisationDetritus
+
+const NPDModels = OceanBioME.Models.NutrientsPlanktonDetritusModels
+const DetritusModels = NPDModels.DetritusModels
+
+struct ExternalContractPlankton end
+
+Oceananigans.Biogeochemistry.required_biogeochemical_tracers(::ExternalContractPlankton) =
```
jagoosw: Dont these need to be imported to extend?

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/utils.jl
nanophyto: It is meant to be a generic fallback for external plankton components such that non plankton groups don't accidentally enter conservation checks.

Maybe this is clearer?

```
# Fallback for externally defined plankton components.
function group_element_tracers(group, bgc, val_element)
    if group !== bgc.plankton
        return NamedTuple()
    end
```

test/test_external_plankton_interface.jl
nanophyto: Not strictly, using the fully qualified function name is also sufficient to extend it. But an explicit import is more consistent with the other OceanBioME tests.

src/Models/AdvectedPopulations/NutrientsPlanktonDetritus/utils.jl
jagoosw: Oh sorry I understand is this a fallback for these method definitions
https://github.com/OceanBioME/OceanBioME.jl/pull/393/changes/BASE..3e91e4744455689c3ae7bbd3e537d53755e5eb4d#diff-c0c00769618f3f9f6c87885cc90a86ec0d107b12e800c2f088dbb205d8038d4dL51

nanophyto: Yes, I have now changed the syntax for clarity

nanophyto:
@jagoosw it seems the tests now pass. I think I addressed your suggestions but let me know if you spot any other issues with this PR.

jagoosw (approved):
This looks great, thank you for implementing!

jagoosw:
Can I merge?

nanophyto:
> Can I merge?

Yes, thanks!

jagoosw:
I just realised that this doesn't bump that patch number, do you want me to do that so we can make a release with these changes?

nanophyto:
> I just realised that this doesn't bump that patch number, do you want me to do that so we can make a release with these changes?

Yes I think a release would make sense since it introduces some minor breaking changes?

PR #396, opened 2026-08-21
(0.18.2) Prepares for MARBL

jagoosw:
This PR includes a few internal changes we need to prepare for MARBL:
- [x] Add MARBL's DRTSafeSolve algorithm for exact reproduction of carbon chemistry solves
- [x] Add possibility to output CO2 concentration (rather than pco2 or pH) from carbon chemistry solver for diagnostics
- [x] Allow sediment to access all fields rather than just prognostic fields
- [x] Update the gas constant value to match MARBL (83.14472 to 83.1451)
- [x] Change default sulfate concentration from `0.14 / 96.06 * S / 1.80655` to `0.14 / 96.062 * S / 1.80655`


johnryantaylor:
I made a small change to the carbon chemistry docstring.  Otherwise this looks good to me.

lgloege (approved):
looks good

PR #399, opened 2026-08-24
(0.19.0) Changes for NumericalEarth coupling + gas exchange changes #411

jagoosw:
This PR makes a few changes we need for automatic NumericalEarth setup:

- [x] Introduces an PAR abstract type so we can set up things effecting all models more easily (and introduces some helper functions)
- [x] Introduces `PARFromShortwave` which has a `surface_shortwave` (that can be a number/field/function), and a PAR fraction
- [x] Introduces an abstract type for inorganic carbon models to pass around the number of replicates generically
- [x] Moves wind speed inside the transfer velocity to allow transfer velocities in general to depend on more things
- [x] Allows `surface_value` to interpolate field time series

Proceeds https://github.com/NumericalEarth/NumericalEarth.jl/pull/572

This PR is breaking in the gas exchange API as it moves the wind speed into the transfer velocity rather than being a top level property of the gas exchange. This is only breaking if you construct directly rather than using the co2 or oxygen convenience constructors (which I guess people probably don't).

Via #411: In the API the breaking changes are dropping first_virial_coefficient, cross_virial_coefficient, and air_pressure from CarbonDioxideConcentration, and solubility from the transfer velocity wrappers (they are now plain transfer velocities). The value of CO2 fluxes has also changed.


src/Models/GasExchange/gas_solubility.jl
```diff
@@ -23,6 +23,7 @@ end
 @inline surface_value(gs::PartiallySolubleGas, i, j, grid, clock, model_fields) = 
     surface_value(gs.air_concentration, i, j, grid, clock) * surface_value(gs.solubility, i, j, grid, clock, model_fields)
 
+# this isn't used anywhere?
```
johnryantaylor: why is this comment here?

src/Models/GasExchange/gas_transfer_velocity.jl
```diff
@@ -27,11 +28,25 @@ struct SchmidtScaledTransferVelocity{KB, SC, SO}
...
+    T = @inbounds model_fields.T[i, j, grid.Nz]
+    S = @inbounds model_fields.S[i, j, grid.Nz]
+
+    # maybe we want to generalise these?
+    Sc = k.schmidt_number(T)
+    ζ  = k.solubility(T, S) # this is maybe a bad name, this is really the conversion from gas to concentration I think
+
+    k₀ = surface_value(k.base_transfer_velocity, i, j, grid, clock, model_fields)
+
+    return k₀ * sqrt(Sc / convert(eltype(model_fields.T), 660)) * ζ
```
johnryantaylor: The docstring in gas_transfer_velcoity.jl implies that this should be k*sqrt(660/Sc).  Do you know which one is correct?

PR #401, opened 2026-08-28
correct nitrogen fixation parameters in nitrate_ammonia.jl

johnryantaylor:
I was doing a test of nitrogen fixation which I realized never turns on in PISCES.  This was happening due to a unit mismatch between the growth rate (s^-1) and the limiting rate (copied from NEMO in days^-1.  This edit fixes that.


src/Models/AdvectedPopulations/PISCES/nitrogen/nitrate_ammonia.jl
```diff
@@ -9,7 +9,7 @@ phytoplankton. Additionally waste produces ammonia through various means.
 """
 @kwdef struct NitrateAmmonia{FT}
               maximum_nitrification_rate :: FT = 0.05  / day # 1 / s
-                   maximum_fixation_rate :: FT = 0.013 / day # mmol N / m³ (maybe shouldn't be a rate)
+                   maximum_fixation_rate :: FT = 0.013       # mmol N / m³
```
jagoosw: Is this meant to be changed too? I checked the paper and its listed as per day and it appears in the NH4 tendency directly?

johnryantaylor: This was intentional.  growth_requirement has units of 1/time.  In the code before this edit there were two terms with units of 1/time (and the time units of mu and 2.15 didn't match).  Another approach would be to keep maximum_fixation_rate in units of 1 /day and make growth_requirement non-dimensional.  Maybe that would be more consistent with the Aumont paper.

jagoosw: Oh yes I understand now. I think this has units of mol N/mol C then right?

johnryantaylor: I changed this so that the rate is still in units of day^-1 and non-dimensionalized the growth rate below.  That is probably less confusing (still somewhat confusing, but that is more about the PISCES formula, I think)

PR #405, opened 2026-09-04
(0.18.6) Introduces abstract light which supports interface computation

jagoosw:
This is a precursor to #398 and creates an abstract light attenuation model type. The abstract type supports both centered and interface light (which computes the exponentially weighted mean at centers). In the existing two band and prescribed attenuation models you compute on the interfaces by setting the kwarg `interface_field` to a `ZFaceField` (or actually anything that isn't `nothing` but that almost definitely won't work).

This now requires light models to specify the number of bands in the abstract type (which I guess could be a user specified amount, but that isn't implemented), have a parameter `field`, and optionally `interface_field`. New light models should define a method for `attenuation`, if only specifying a single band:
```julia
attenuation(i, j, k, grid, light_model, clock, Chl)
```
if multiple:
```julia
attenuation(i, j, k, grid, light_model, clock, Chl, ::Val{n})
```
with `n` for each band. Models may also specify `weight(::AbstractLight{N}, ::Val{N})` for each band (fallback to even weighting).

I'll check but I'm also pretty sure the interface + average to centers is actually more accurate/correct.

This does not support the Morel multi band model because that keeps different bands explicitly (as per PISCES), where as the others only track a single PAR which can have multiple bands internally.


jagoosw:
Would anyone mind if I change the default to compute at interfaces and average to center (but not store the interfaces)? I think this is actually what we want because in the finite volume framework we want the phytoplankton to depend on the cell average PAR not the PAR at the cell center?

src/Light/abstract_light.jl
```diff
@@ -0,0 +1,135 @@
...
+@kernel function integrate_light_attenuation!(la::AbstractLight{1, Nothing}, 
+                                              PAR, grid, clock, Chl, surface_PAR)
+
+    i, j = @index(Global, NTuple)
+
+    PARᵢ = getbc(surface_PAR, i, j, grid, clock, Chl)
+
+    @inbounds for k in grid.Nz:-1:1
+        eᵏᵈᶻ = face_to_face_attenuation(i, j, k, grid, la, clock, Chl)
+        PAR[i, j, k] = - PARᵢ * (1 - eᵏᵈᶻ)/log(eᵏᵈᶻ)
```
johnryantaylor, line 17: Does this NaN when attenuation is zero?  Maybe we should add a small epsilon to prevent that.  I think the code had this issue before too, but we could fix it.

johnryantaylor:
This looks good to me (note my comment above about possibly including a tol to handle very small attenuation values).  I see that the GPU tests are failing.  I don't know if that is new or if it is the issue with buildkite that we have been having recently.

jagoosw:
> This looks good to me (note my comment above about possibly including a tol to handle very small attenuation values). I see that the GPU tests are failing. I don't know if that is new or if it is the issue with buildkite that we have been having recently.

It was a problem with an adapt, fixed now!

For the $e^{-k\Delta z}\to0$ case, I'm not sure it would ever happen because we calculate that term directly, so unless $K$ or $\Delta z$ are `Inf`, $\log\left(e^{-k\Delta z}\right)$ will be finite I think?

johnryantaylor:
> > This looks good to me (note my comment above about possibly including a tol to handle very small attenuation values). I see that the GPU tests are failing. I don't know if that is new or if it is the issue with buildkite that we have been having recently.
> 
> It was a problem with an adapt, fixed now!
> 
> For the e − k Δ z → 0 case, I'm not sure it would ever happen because we calculate that term directly, so unless K or Δ z are `Inf`, log ⁡ ( e − k Δ z ) will be finite I think?

Oh, yeah, you're right.  I saw the function name face_to_face_attentuation and I thought that it was the light attenuation which could be near zero.  What you are calculating is actually the exp() of the attentuation which shouldn't be zero.  Maybe worth changing the function name to exp_face_to_face_attenuation to clarify, though?

PR #410, opened 2026-09-09
(0.19.0b) Improves gas exchange + adds MARBL's oxygen solubility

jagoosw:
This PR changes the water side computation to a use the aquious concentration ([ $CO_2\text{(aq)}$ ]) rather than figacity ([ $fCO_2$ ]) for consistency with MARBL and the OCMIP protocol. The *flux* is now compared to MARBL computed values which is an improvement on the previous testing which just included the different parts.

This PR also introduces `GarciaGordonOxygenSaturation` which goes in the `air_concentration` slot for oxygen and computes the saturation directly from temperature and salinity as per MARBL etc. 

In the API the breaking changes are dropping `first_virial_coefficient`, `cross_virial_coefficient`, and `air_pressure` from `CarbonDioxideConcentration`, and `solubility` from the transfer velocity wrappers (they are now plain transfer velocities). The value of $CO_2$ fluxes has also changed.


jagoosw:
I'm going to merge after #399 then release both together

jagoosw:
Moving to #411 to make just one breaking change

PR #415, opened 2026-09-14
(0.18.9) Fix NaN in two-band PAR below the red band's underflow depth

### LINKED ISSUE #413: Two-band PAR cell average returns `NaN` below ~3300 m (~460 m on `Float32`)
Since [#405](https://github.com/OceanBioME/OceanBioME.jl/pull/405), `TwoBandPhotosyntheticallyActiveRadiation` is an [`AbstractLight{2}`](https://github.com/OceanBioME/OceanBioME.jl/blob/358ed3b642cdfbb38394fffef1fad78353f5ca5a/src/Light/2band.jl#L1) and runs through the shared multi-band kernel in [`abstract_light.jl`](https://github.com/OceanBioME/OceanBioME.jl/blob/358ed3b642cdfbb38394fffef1fad78353f5ca5a/src/Light/abstract_light.jl#L44-L60).

That kernel accumulates each band's transmittance `K[n]` down the column and then builds the cell average from the **ratio of successive accumulated products**, `K_next[n] / K[n]`. `K[n]` decays like `exp(-kₙ · depth)`, so once it underflows to exactly `0.0` the ratio is `0/0 = NaN`, and PAR is `NaN` for the rest of the column. `NaN` PAR makes every biogeochemical tracer non-finite within a time step or two.

With the default [`water_red_attenuation = 0.225 /m`](https://github.com/OceanBioME/OceanBioME.jl/blob/358ed3b642cdfbb38394fffef1fad78353f5ca5a/src/Light/2band.jl#L82) the red band underflows at `745 / 0.225 ≈ 3310 m` on a `Float64` grid and at `103 / 0.225 ≈ 460 m` on a `Float32` grid.

I think there should be a way to rewrite the code to just avoid the underflow.

## MWE

```julia
using Oceananigans, OceanBioME

for (FT, depth) in ((Float64, -3000), (Float64, -4870), (Float32, -400), (Float32, -600))
    grid = RectilinearGrid(CPU(), FT; size = (4, 4, 60), x = (0, 1e4), y = (0, 1e4), z = (depth, 0))
    biogeochemistry = LOBSTER(grid) # default light model is two-band PAR
    model = HydrostaticFreeSurfaceModel(grid; biogeochemistry, tracers = (:T, :S), buoyancy = nothing)
    Oceananigans.TimeSteppers.update_state!(model)
    PAR = interior(biogeochemistry.light_attenuation.field)
    println(FT, ", ", -depth, " m: ", count(isnan, PAR), " NaN PAR cells of ", length(PAR))
end
```

```
Float64, 3000 m: 0 NaN PAR cells of 960
Float64, 4870 m: 304 NaN PAR cells of 960
Float32, 400 m: 0 NaN PAR cells of 960
Float32, 600 m: 208 NaN PAR cells of 960
```

ali-ramadhan:
This PR resolves #413 by:
1. modifying `exponential_face_to_face_attenuation` to return just the per-cell transmittances `t = exp(-Δz * attenuation)`, no longer multiplied by the cumulative attenuation, so it means the same thing as the single-band method of the same name.
2. adding a new `attenuate(K, t, la)` that accumulates `K .* t`.
3. modifying `total_cell_average(K, t, la)` to use `(1 - t) / log(t)` directly instead of `(1 - K_next/K) / log(K_next/K)`.

So now an underflowed band contributes exactly zero and PAR falls back to the remaining bands.


jagoosw:
~In the finite region does this give the exact same PAR?~
I checked and it does, I'll bump the patch version and merge this

jagoosw (approved):
Thank you for finding this issue!

ali-ramadhan:
Awesome, thanks @jagoosw!

PR #422, opened 2026-09-18
Adds MITgcm DIC package iron scavenging and dust remineralisation parameterisation

jagoosw:
This PR allows us to fully replicate the MITgcm DIC package in this [repo](https://github.com/MITgcm/MITgcm/tree/master/pkg/dic) described [here](https://mitgcm.readthedocs.io/en/latest/phys_pkgs/dic.html).

It is an implicit biology model which tracks only $PO_4$, DOP, DIC, Alk, and optionally Fe and $O_2$. We already had the implicit biology model so we just needed to add the specific iron chemistry (some scavenging which is lost, implicitly to permanent sediment storage), and their parameterisation for the supply of iron from dust.


jagoosw:
I don't like the API claude has created here so doing some more work on it before merging

src/Utils/iron_dust_deposition.jl
```diff
@@ -0,0 +1,65 @@
+using Oceananigans.BoundaryConditions: FluxBoundaryCondition, getbc
+
+import Adapt: adapt_structure, adapt
+import Base: summary, show
+
+"""
```
jagoosw, line 8: This whole thing is incorrect

jagoosw:
image
Simple iron limited case +/- dust deposition

PR #424, opened 2026-09-19
Adds implicit particle sinking

jagoosw:
As per MARBLs implementation, this PR adds implicit particle sinking where, instead of an explicit advection, sinking happens quasi-instantaniously with the sinking _flux_, $F$, following the equation:
```math
\frac{\partial F}{\partial z} = \Pi - \frac{F}{\ell},
```
where $\ell$ is a sinking length scale (which can vary with depth etc. like in MARBL and $\Pi$ is the local production. So remineralisation is:
```math
R = \Pi + \frac{F_{\text{in}} - F_{\text{out}}}{\Delta z}.
```
This is solved by backwards Euler from the surface:
```math
F_{\text{out}} = \frac{F_{\text{in}} + \Pi \cdot \Delta z}{1 + \Delta z / \ell},
```
with $F$ on cell faces.

This can replace explicit sinking in all existing detritus models, I think it will mainly be of use for the low resolution global model where the timestep could be $1.5$hours and the most fine vertical grid spacing be $1.5$m making the maximum sinking speed be 24m/day (at CFL=1) which is too slow.

To use this you pass a NamedTuple of tracer names and length as `dissolution_lengths` instead of `sinking_speeds` (which should be set to `nothing` for clarity.


jagoosw:
https://github.com/user-attachments/assets/bc755e56-97f5-49d6-bc17-5aa0ce53a597

In the top left N+D is plotted for the explicit case but N for the implicit because D is instantly remineralised. In the bottom right the remineralisation rate is higher in the explicit case because some of the detritus is mixed back up and remineralised there, where as in the implicit case it is only "transported" down

